PR #26, opened 2020-07-25
Bug Fix and Typo Fix

Jobsecond:
1. Fix the bug that causes the application to crash when canceling the "Import Midi..." OpenFileDialog.
2. Fix a typo in `UTrack.cs`.


stakira (approved):
Thanks!

PR #49, opened 2021-08-07
Singer character.txt parsing fixes

messmerd:
Some Utau singers, such as the ones at [this site](https://marginalsynth.wixsite.com/utau/) use ": " as a delimiter in their character.txt files rather than "=". For this reason, singers from that site (and potentially other sites) fail to import due to character.txt parsing issues. 

To fix this, I made OpenUtau fall back on attempting to parse using a ": " delimiter if "=" does not work. Now singers from that site load without issue.

I also made it read the "created by" key in character.txt, and set the author information based on that. Now character.txt files that have a "created by" key but no "author" key can still have their author information loaded.

I have successfully tested my changes with [this singer](https://marginalsynth.wixsite.com/utau/sayuri).

### DIFF
--- a/OpenUtau/Classic/VoicebankInstaller.cs
+++ b/OpenUtau/Classic/VoicebankInstaller.cs
@@ -96,6 +96,8 @@ namespace OpenUtau.Classic {
                             while (!streamReader.EndOfStream) {
                                 string line = streamReader.ReadLine().Trim();
                                 var s = line.Split(new char[] { '=' }, StringSplitOptions.RemoveEmptyEntries);
+                                if (s.Length != 2)
+                                    s = line.Split(new string[] { ": " }, StringSplitOptions.RemoveEmptyEntries);
                                 if (s.Length == 2) {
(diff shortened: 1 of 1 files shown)


OpenUtau/Classic/VoicebankInstaller.cs
```diff
@@ -96,6 +96,8 @@ class FileLoc {
                             while (!streamReader.EndOfStream) {
                                 string line = streamReader.ReadLine().Trim();
                                 var s = line.Split(new char[] { '=' }, StringSplitOptions.RemoveEmptyEntries);
+                                if (s.Length != 2)
+                                    s = line.Split(new string[] { ": " }, StringSplitOptions.RemoveEmptyEntries);
```
stakira, line 100: Thanks for the PR! Could you make this split using char ':', then call string.Trim() on each part? Otherwise it looks good to me.

messmerd: Ok, now the parts are trimmed after the line is split.

PR #56, opened 2021-09-26
JA CVVC phonemizer using hardcoded settings

adlez27:


OpenUtau.Plugin.Builtin/JapaneseCVVCPhonemizer.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                // Current note is VV
+                if (vowelLookup.TryGetValue(prevUnicode.Last(), out var vow)) {
+                    currentLyric = $"{vow} {currentLyric}";
+                }
+            }
+
+            if (nextNeighbour != null) {
+                var nextUnicode = ToUnicodeElements(nextNeighbour?.lyric);
+                var nextLyric = string.Join("", nextUnicode);
```
stakira, line 99: This too.

stakira (commented):
Fantastic work!

OpenUtau.Plugin.Builtin/JapaneseCVVCPhonemizer.cs
adlez27: Next lyric is used to retrieve OTO value to determine VC length

PR #64, opened 2021-09-28
Strings.nl.NL.xaml

RhelaRazer:
Add Dutch translation


stakira:
Added in 2bfa5030 instead due to the conflict.

PR #146, opened 2021-11-10
bring back Modulation

HeidenBZR:
i suppose it was removed long ago when there were no expressions?


stakira:
Thanks for figuring it out. I have been wondering what's the cause.

PR #148, opened 2021-11-11
do not parse file system for expired files on each render

HeidenBZR:
we don't suspect file will expire just while the user is using the program, if the expiration time is 7 days


stakira:
Good point. However, RenderEngine constructor is called just as often as Render**() methods.

PR #157, opened 2021-11-14
Heidenbzr async parallel dictionary

HeidenBZR:


OpenUtau.Core/Render/RenderEngine.cs
```diff
@@ -78,19 +78,20 @@ public class Progress {
...
-                    MaxDegreeOfParallelism = threads
-                }, body: item => {
-                    if (source.Token.IsCancellationRequested) {
-                        return;
-                    }
-                    item.progress = progress;
-                    Resample(item);
-                });
-                progress.Clear();
+                if (items.Count > 0) {
```
HeidenBZR, line 82: "phonemizer is initializing" was overlapped by this "Clear" even if there was no render items

OpenUtau/ViewModels/NotesViewModel.cs
```diff
@@ -509,6 +509,10 @@ public class NotesViewModel : ViewModelBase, ICmdSubscriber {
                     if (focusNote.part == Part) {
                         FocusNote(focusNote.note);
                     }
+                } else if (cmd is ValidateProjectNotification) {
+                    Project.Validate();
```
stakira, line 513: Please keep Project.Validate() call inside DocManager.

OpenUtau.Core/Render/RenderEngine.cs
stakira: Please prefer returning early, to reduce nesting i.e.
```
if (item.Count == 0) {
  return;
}
```

OpenUtau.Core/Api/G2pDictionary.cs
```diff
@@ -30,13 +30,20 @@ class TrieNode {
...
                         .ToList()
                         .ForEach(parts => builder.AddSymbol(parts[0], parts[1]));
-                    Core.Api.Resources.cmudict_0_7b.Split('\n')
-                        .Where(line => !line.StartsWith(";;;"))
-                        .Select(line => line.Trim().ToLowerInvariant())
-                        .Select(line => line.Split(new string[] { "  " }, StringSplitOptions.None))
-                        .Where(parts => parts.Length == 2)
-                        .ToList()
-                        .ForEach(parts => builder.AddEntry(parts[0], parts[1].Split().Select(symbol => RemoveTailDigits(symbol))));
+                    Core.Api.Resources.cmudict_0_7b.Split('\n').AsParallel().ForAll(line => {
```
stakira, line 33: Parallelizing a lot of small tasks is overkill and may actually slow things down.
A single thread as the original is not too bad after you made it async. The builder is not thread-safe either.

PR #210, opened 2022-01-04
Update Vietnamese VCV Phonemizer

janikyou:


janikyou:
I'm sorry, I don't know more complicated algorithms that's why it's so big

PR #212, opened 2022-01-06
Update Vietnamese VCV Phonemizer

janikyou:
I don't know complicated algorithms, it's quite laggy when I write the first note, after that it's normal


stakira:
Please make the phonemizer load a dictionary file instead of hard coding the entire dictionary in code.

PR #236, opened 2022-02-23
Port-For-Linux

mhbalthasar:
Optimize Compatibility of Linux distributions (libdl). Considering the different versions of GCC, the portaudio library of the system is preferred. And added arm64 version of Linux file, adapted to Pi3 and Huawei&Pythom CPU


stakira:
There will be native linux arm64 build soon

PR #247, opened 2022-03-11
French Phonemizer

oriswalft:
APRASing phonemizer for French voicebank on OpenUtau


stakira:
Thanks a lot for contributing. However, I don't think it will work as intended. This PR seem to copy the pt-br model for fr, which would not work. Other issues are duplicated code and redundant files. I'm creating #259 instead.

PR #286, opened 2022-03-30
Add an option to move back cursor on pausing

### LINKED ISSUE #114: Recovery of playback time
I hope I can go back to the first playback time I ran after the play.

shniubobo:
Closes #114

---

One thing to notice is that `PlaybackManager` is at present not good at seeking to a new position. User experience would be better if the `Task` or the `RenderEngine` created in `PlaybackManager.Render` could be reused after seeking. I'll create a seperate issue later to track this.


shniubobo:
Because of the issue template created in 1a3c313c4c5866784e15eac94131e26935623c41, I'm not able to open an issue other than a bug report 😕

stakira:
Thanks a lot for the contributions!

PR #296, opened 2022-04-06
Update Strings.ko-KR.axaml

RyuuseiKang:
번역되지 않아있던 한국어를 수정하였습니다.

Fixed Korean that was not translated.


Cardroid:
I think the other expression is correct.
Thank you for applying for Korean. 👍

PR #298, opened 2022-04-07
Implement alias override UI

shniubobo:
Only tested with JP VCV. Please help test with other phonemizers.


stakira:
The space key doesn't work in AliasBox. Also other deactivation (e.g., right click piano roll to close it) issues. See all lyricBox references in PianoRollWindow.axaml.cs.

This leads to the question: is it possible to reuse LyricBox?

shniubobo:
> The space key doesn't work in AliasBox. Also other deactivation (e.g., right click piano roll to close it) issues. See all lyricBox references in PianoRollWindow.axaml.cs.

My bad. I had noticed the deactivation issue when testing but later forgot to fix it. I didn't notice the space key one though. Thanks for pointing out those issues.

> This leads to the question: is it possible to reuse LyricBox?

This depends on how you define "reuse LyricBox".

If that means having two different controls that share the suggestion and input parts, then I don't think it's necessary, though it's possible. This would result in three `UserControl`s (LyricBox, AliasBox, and the suggestion and input part as a seperate control). There would still be a lot of code duplication (they still have their own view models and code-behind), while the complexity would have increased drastically at the same time.

But if it means having only LyricBox, and having it behave differently according to its use, then I think that is possible as well as desirable. Code duplication would be reduced a lot, and the complexity wouldn't increase much.

As an aside, by "having it behave differently according to its use", I mean calling `LyricBox.Show` with an additional argument `bool isAliasBox`, and according to this argument, let LyricBox:

1. Show either the tab shift-tab tips or the "Override Alias" title
2. On pressing tab, either switch to the adjacent notes or do nothing
3. After selecting the lyric / alias, execute either `ChangeNoteLyricCommand` or `ChangePhonemeAliasCommand`

Anyway, I will try implementing the latter, and let you know when it's finished. Also, please let me know if I need to squash this commit with [e27e357](https://github.com/stakira/OpenUtau/pull/298/commits/e27e3576d00b28e07ef0eb7f34c2245d42a42de5).

stakira:
Good job!

PR #315, opened 2022-04-20
[Spanish Phonemizer] Fixed consonant transition issue & added new functions 

lottev1991:
* Consonant transition issues got fixed.
* Now supports VCV with more than one consonant if the voicebank has it.
* Supports "consonant VCV", e.g. [n da], [l ba] etc. if the voicebank has it.
* Consonant transition adjustment

**TODO:** make VCV with "ny" work automatically as a substitute for "nh" if the voicebank has it, instead of always falling back to CVVC.


lottev1991:
Found an error, my apologies.

PR #316, opened 2022-04-20
[Spanish Phonemizer] Fixed transition issues & added new functions 

lottev1991:
* Consonant transition issues got fixed.
* Now supports VCV with more than one consonant if the voicebank has it.
* Supports "consonant VCV", e.g. [n da], [l ba] etc. if the voicebank has it.
* Consonant length adjustment.
* "ny" is now a fallback for "nh" if the voicebank doesn't have the latter (CVVC-only).
* Use seseo ("s" instead of "z") when a voicebank doesn't have "z" (might be CVVC-only).

**TODO:** 
* Make VCV with "ny" work automatically as a substitute for "nh" if the voicebank has it, instead of always going for CVVC.
* Make ValidateAlias for using "i" instead of "y" and "u" instead of "w" work properly if the voicebank has them.


lottev1991:
Found another error, apologies again.

PR #320, opened 2022-04-23
Add note defaults & presets

Fyne-Q:
Added a window for default notes settings - lyric, portamento and vibrato.
- Portamento settings include start and length;
- Vibrato settings include length, period, depth, fade in and out, and shift, as well as toggle for automatic vibrato and minimum note duration for adding it;
- Portamento and vibrato have presets, which user can add or remove;
- Default settings and presets are saved in a file separate from preferences;
- Batch edit previously named "Reset vibratos" is now named "Clear vibratos" and current "Reset..." applies all the default values;
Small bug which I wasn't able to fix: preset dropdown lists don't update after adding or removing presets, the window needs to be reopened for changes to be seen.

### DIFF
--- a/OpenUtau.Core/Commands/ExpCommands.cs
+++ b/OpenUtau.Core/Commands/ExpCommands.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 
 using OpenUtau.Core.Ustx;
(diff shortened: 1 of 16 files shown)


stakira:
🎉🎉🎉 This is so great. I'm sure a lot of people will appreciate it.

PR #346, opened 2022-05-08
EN VCCV (Cz's) Phonemizer 

mmemim:
This is a temporary workaround for the lack of phonemizer for Cz's VCCV method. Since it uses an arpabet dictionary, it lacks some vowels right now but it is fully working. _Known issues:_
- There may be some more VC exceptions to add. 
- The phonemizer defaults to [V C] for parsing, maybe it should be [VC-]. Feedback welcome, please test it yourself! 
- CC conflicts with "zh/z+h" "th/t+h" etc.

**TODO:** 
- Make VC exceptions cleaner and sturdier (maybe use GetSymbols())
- Work out a solution for additional vowels and sounds (dd, h/hh)

Feel free to contribute. :) 
The Lyric Parser plugin is still a sturdier alternative.

### DIFF
--- a/OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
+++ b/OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
@@ -245,14 +245,16 @@ namespace OpenUtau.Plugin.Builtin {
             }
 
 
-
-            if (!HasOto(basePhoneme, syllable.tone)) {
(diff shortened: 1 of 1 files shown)


mmemim:
Currently adding some new logic with "prevWordConsonantsCount"

mmemim:
Didn't add Length by Oto in the end because it didn't sound good.

PR #434, opened 2022-08-25
[Korean CVC Phonemizer] Fix "ㄹ" sandhi rule

lottev1991:
This fixes the "ㄹ" sandhi rule in notes with a null initial ("ㅇ").


lottev1991:
never mind

PR #453, opened 2022-09-07
Add ghost notes feature (show notes in other tracks in piano roll)

lselden:
This PR adds a "Ghost Notes" feature to OpenUtau, that shows the notes of other tracks when viewing a part in the Piano Roll. This is similar to the "Ghost Notes" feature in FL Studio, or Ableton's multi-clip view.

* Added preferences option for enabling/disabling feature
* Handles part changes/moves

ghost-notes

### TODO
[ ] Review naming / language - I'm not married to "Ghost Notes"
[ ] Review styling choices (currently uses existing "NeutralAccentBrushSemi" brush color, and display slightly thinner)
[ ] Add additional translation strings
[ ] Choose default option (enabled/disabled)
[ ] Handle initial load when preference not set *(I'm not familiar with this behavior)*

### DIFF
--- a/OpenUtau/Controls/NotesCanvas.cs
+++ b/OpenUtau/Controls/NotesCanvas.cs
@@ -106,7 +106,7 @@ namespace OpenUtau.App.Controls {
         public NotesCanvas() {
             ClipToBounds = true;
             pointGeometry = new EllipseGeometry(new Rect(-2.5, -2.5, 5, 5));
-            showGhostNotes = Convert.ToBoolean(Preferences.Default.ShowGhostNotes);
+
             MessageBus.Current.Listen<NotesRefreshEvent>()
                 .Subscribe(_ => InvalidateVisual());
             MessageBus.Current.Listen<NotesSelectionEvent>()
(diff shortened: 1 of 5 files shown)


stakira:
- "Ghost note" doesn't seem to be confusing so it's fine
- I think it looks good 👍 
- Don't worry about translations, other people can contribute. Before that English is the fallback.
- I think default to enabled is fine.
- The default value in `Preferences.cs` is all it needs. So no worries.

OpenUtau/Controls/NotesCanvas.cs
```diff
@@ -98,9 +100,13 @@ class NotesCanvas : Canvas {
         private HashSet<UNote> selectedNotes = new HashSet<UNote>();
         private Geometry pointGeometry;
 
+        private bool showGhostNotes = true;
+        private List<UPart> otherPartsInView = new List<UPart>();
+
         public NotesCanvas() {
             ClipToBounds = true;
             pointGeometry = new EllipseGeometry(new Rect(-2.5, -2.5, 5, 5));
+            showGhostNotes = Convert.ToBoolean(Preferences.Default.ShowGhostNotes);
```
stakira, line 109: This makes preference change only apply after relaunch. Could we move the check to RefreshGhostNotes()?

lselden: Fixed, and that move does load preferences correctly

stakira:
🎉 🎉 🎉

PR #458, opened 2022-09-12
Update CVVC/VCV Vietnamese Phonemizer

janikyou:


VietnameseCVVCPhonemizer.cs
```diff
@@ -0,0 +1,1395 @@
+﻿using System.Collections.Generic;
```
stakira, line 1: are these two files added by mistake? could you remove?

PR #484, opened 2022-09-21
[JA CVVC Phonemizer] Multiple bug fixes + added function

lottev1991:
- Fixed "double VC" bug.
- Add "dz" + "z" substitute.
- Revert ``を`` back to ``plainVowels`` (the large majority of CVVC banks treat this as a vowel).
- Voice color fix (a proper fallback system for non-vowels and non-initials was not defined, which resulted in the error).
- VV and crossfade vowel tests are now switched, meaning that VVs get priority.

**TODO for future PR:** Create ``を`` fallback? (i.e., if there's no VV note, fallback to VC note)


HitCoder9768:
Looks like this may be good to go IMO

PR #532, opened 2022-10-25
updated EN strings, attempted to add file association

mango-ki:
you can probably get rid of the theme stuff


stakira:
Closing due to in-activity

PR #552, opened 2022-11-17
Export the whole project into a single wav file / 混缩导出为单个wav文件

oxygen-dioxide:
Suggested in https://discord.com/channels/551606189386104834/1024455153220329492/threads/1026533351340920852
image

### DIFF
--- a/OpenUtau.Core/Render/RenderEngine.cs
+++ b/OpenUtau.Core/Render/RenderEngine.cs
@@ -49,7 +49,7 @@ namespace OpenUtau.Core.Render {
             this.startTick = startTick;
         }
 
-        public Tuple<MasterAdapter, List<Fader>> RenderProject(int startTick, TaskScheduler uiScheduler, ref CancellationTokenSource cancellation) {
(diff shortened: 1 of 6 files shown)


OpenUtau.Core/Render/RenderEngine.cs
```diff
@@ -103,6 +103,51 @@ class RenderEngine {
             return Tuple.Create(master, faders);
         }
 
+        public WaveMix RenderMixdown(TaskScheduler uiScheduler, ref CancellationTokenSource cancellation) {
```
stakira, line 106: Isn't this the same as `RenderProject(0, ...`?

oxygen-dioxide:
I've already tried `RenderProject(0, ...`. If I export a wav file before playing, it wouldn't wait until the full song is rendered, resulting in an incomplete audio file. Or maybe I should create a switch on the function.
---
试过了`RenderProject(0, ...`，如果启动OpenUTAU之后不播放就导出，那么导出不会等待渲染，导致音频只有开头部分。或者我可能应该给这个函数加个开关

Issue #386, opened 2022-07-09
Open Utau Accessibility Improvements


dangero2000:
### Acknowledgement

- [X] I have read Getting-Started and FAQ

### 🐛 Describe the bug

I am a screen reader user due to having almost 0 eyesight. Unfortunately, Open Utau is not currently accessible with screen reading software, but I'm hoping we can work together to change that.
I have read the Getting Started and FAQ pages, and from those I have come up with a basic idea of where we can start. There are some keyboard shortcuts for actions already, but not every action is accessible with the keyboard and requires the use of a mouse. Actually, most of the actions pointed out in the FAQ are exclusive to the mouse. I think, before adding speech with a screen reader, this would be an amazing place to start. If Open Utau can be used exclusively with the keyboard, you’re already half way to making the program fully accessible.
If you have any questions for me, please do let me know, and I’ll try to answer them to the best of my abilities. Let’s make the music world more accessible together!

### Explains how to reproduce the bug

N/A

### OS & Version

N/A

### Logs

N/A

lselden:
I've been looking at this, as I myself would like to use the keyboard for navigating OpenUTAU more. I'm not ready to open a pull request, but my initial attempts have been promising - I've gotten basic piano roll navigation and open/closing the edit lyrics window working. While I'm poking at it I want to put down some development notes, hoping someone more familiar with the OpenUtau code would want to comment on best approaches.

## Dev Notes
* Architecturally I'm thinking of creating a global key manager, which looks at the currently active view is and send the appropriate action as necessary, similar to how Visual Studio Code handles key bindings. Would it make sense to put that sort of shared functionality in a new folder/namespace of `OpenUtau`?
* One way to support keyboard navigation as well as mouse nav menu items need to use `Commands` rather than the `Click` event handler. This assumes that the Command/Method is on the ViewModel instead of code-behind (*unless someone knows how to target the code-behind using `Binding`*). This also allows using "Accelerators" - just put an underscore (`_`) before the letter used for triggering the menu item. For example, `"Open"` becomes `"_Open"`. Then a user can press `Alt+F` to open `"_File"`, then `O` for the Open option.
[...]

PR #554, opened 2022-11-17
Piano Roll keyboard control

lselden:
This PR addresses #386 and adds a number of additional keyboard shortcuts to the piano roll view. This is a fairly big PR so I tried to break the changes into smaller commits to *(hopefully)* make reviewing easier. 

### Major Features ###
* navigate through notes using Left+Right. Scrolls view as necessary.
* move/resize notes
* enter key opens Lyric Dialog instead of Lyric Box if multiple notes selected (dialog is not accessible and listens for Enter/Esc)
* playback position control
* view scroll/zoom
* add new note with insert key
* **my favorite** - Set playback position to current note with `Ctrl+F`, great when combined with Left+Right to quickly review a phrase

### Changed Behaviors ###
* Lyric Dialog no longer accepts Tab/Enter keys (I don't see the use case for using them. You can still paste in text with these characters). It also focuses the text input area automatically.
* Pressing Tab/Shift+Tab in the lyric box changes the current selection accordingly
* Undoing a note deletion re-selects the deleted notes

### Question ###
Where should the keyboard shortcuts be documented? They could go in the "Tips" popup I guess, but it may get kinda big. Easiest option would be to put in the [Wiki](https://github.com/stakira/OpenUtau/wiki)

### DIFF
--- a/OpenUtau.Core/Commands/NoteCommands.cs
+++ b/OpenUtau.Core/Commands/NoteCommands.cs
@@ -6,7 +6,7 @@ using OpenUtau.Core.Ustx;
 
 namespace OpenUtau.Core {
     public abstract class NoteCommand : UCommand {
-        protected readonly UNote[] Notes;
+        public readonly UNote[] Notes;
         public readonly UVoicePart Part;
         public override ValidateOptions ValidateOptions => new ValidateOptions {
             SkipTiming = true,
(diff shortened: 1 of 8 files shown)


stakira:
Great job!
Yes, feel free to create a new page in the Wiki.

lselden:
Added wiki page [here](https://github.com/stakira/OpenUtau/wiki/Keyboard-Shortcuts)

dangero2000:
This is awesome! Thank you for adding these keyboard shortcuts! I'm sorry I haven't been able to help much with this; life gets pretty busy nowadays, and I confess I'm not the greatest at coming up with decent features. That's why I've been fairly quiet during the last few months; I confess I'm not really confident in how to help, and I'm worried my efforts would just confuse people further. Thanks to you though, we're one step closer to making Open Utau accessible for everyone! :-)

PR #565, opened 2022-12-03
Update Strings.id-ID.axaml

pyorririn:


stakira:
You need to uncomment the changed lines, otherwise it will have no effect.

PR #575, opened 2022-12-21
Fix null lyric on USTX load

CarpetBook:
The English word "null" cannot be used in lyrics when saved as a USTX file. The YAML parser reads a literal `null` on load and fails. [Discord bug report](https://discord.com/channels/551606189386104834/801525671775043625/1054981543170428928)

- Surround UNote lyrics with quotes `" "` to force YAML to serialize literal strings
- Remove quotes from lyric on load
- Literal `null` lyrics convert to default lyric on load to prevent loading null lyric
- Add `AfterSave` to UPart, make UProject execute `AfterSave` to remove multiple quotes on consecutive saves

Admittedly this is probably clunky and not perfect, but it's better than failing to load because of a YAML quirk

### DIFF
--- a/OpenUtau.Core/Ustx/UNote.cs
+++ b/OpenUtau.Core/Ustx/UNote.cs
@@ -65,6 +65,8 @@ namespace OpenUtau.Core.Ustx {
                     exp.descriptor = descriptor;
                 }
             }
+            if (lyric == null) lyric = NotePresets.Default.DefaultLyric;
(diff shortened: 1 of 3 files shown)


oxygen-dioxide:
"null" string can not only exist in lyrics but also in other strings, such as voicebank names, part names. so this should be global for yaml serializer. You can use WithQuotingNecessaryStrings in [OpenUtau.Core/Util/Yaml.cs](https://github.com/stakira/OpenUtau/blob/master/OpenUtau.Core/Util/Yaml.cs)

PR #586, opened 2023-01-25
expression line draw

lennyservant:
adds expression line drawing

https://user-images.githubusercontent.com/45115987/215006806-863464f7-00e7-4b21-9c06-969dae92b7a5.mp4


lennyservant:
[shift key implementation](https://github.com/stakira/OpenUtau/pull/586/commits/0359d014b9872b75ffb8e286376633c55fe5f024) feels like it could cause problems in the future, is there another way to do it?

PR #618, opened 2023-02-25
String sort localization / 字符串排序本地化

oxygen-dioxide:
After this change, users will be able to sort their voicebanks in a certain language, such as pinyin (zh-cn). This would be very convenient if the user have many voicebanks installed.
The sorting order can be different from the UI language. For example, If a Chinese user have a lot of Japanese voicebanks installed, he may want to sort his voicebanks in Japanese while seeing OpenUtau's UI in Chinese.


oxygen-dioxide:
Here is microsoft's official documentation on the neccessarity of localized string sorting: https://learn.microsoft.com/en-us/globalization/locale/sorting-and-string-comparison

PR #621, opened 2023-02-28
A (semi) working implementation for custom themes. [WIP/DO NOT MERGE]

Astel123457:
Right now, it's just bug fixes and tweaks

KNOWN BUGS:
- Theme names don't show up when changing themes, however they can be changes when clicked on.
- It is possible that languages may interfere with theme switching, so I will do some testing with that


Astel123457 (commented):
fixed the oopsie

PR #642, opened 2023-04-04
Fix issue with the Stoplight buttons on OSX

Astel123457:


stakira:
The issue has been fixed. Thanks still!

PR #655, opened 2023-04-13
Add "Japanese presamp Phonemizer" JP VCV & CVVC

maiko3tattun:
* Added "Japanese presamp Phonemizer"
* Fixed "Japanese CVVC Phonemizer", bug fix on VC length.


oxygen-dioxide:
Both "JA VCV & CVVC" and "ZH CVVC" are based on presamp.ini. How about making a base presamp phonemizer, and make ZH CVVC and JA CVVC two subclasses of it? Like how EN VCCV and FR CVVC are based on SyllableBasedPhonemizer.

PR #656, opened 2023-04-14
loading of Append.maps for multi-prefix maps

maiko3tattun:
Improved loading of Append.maps for multi-prefix map.
The (Append).maps in the prefix folder is used by presamp and some of the classic UTAU plugins, which are added to subbanks.
(Among the changes, #655 is also included)


maiko3tattun:
- Implementation of the Presamp class
- Japanese presamp Phonemizer now reads presamp.ini
- Added TryGetLines() to Ini class

The presamp class holds all the data in presamp.ini, but it is not necessary for the phonemizer to use all the data.
I created it in Classic for now.

maiko3tattun:
- Add JaPresampTest

PR #661, opened 2023-04-18
Pitch Baking

oxygen-dioxide:
An editing macro that converts PITD to pitch points is added.

### DIFF
--- a/OpenUtau.Core/Editing/NoteBatchEdits.cs
+++ b/OpenUtau.Core/Editing/NoteBatchEdits.cs
@@ -305,6 +305,13 @@ namespace OpenUtau.Core.Editing {
         }
 
         //reference: https://github.com/sdercolin/utaformatix3/blob/0f026f7024386ca8362972043c3471c6f2ac9859/src/main/kotlin/process/RdpSimplification.kt#L43
+        /*
+        * The Ramer–Douglas–Peucker algorithm is a line simplification algorithm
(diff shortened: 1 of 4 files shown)


oxygen-dioxide:
Current state: pitch point reducing strategy is adopted from https://github.com/oatsu-gh/ENUNU/blob/main/synthesis/extensions/f0_feedbacker.py . Only extreme points are preserved, which will lose many features when the pitchline is going up or down, but keep too much points when the pitchline is nearly horizontal. 

Todo:
- improve the pitch point reducing strategy. (feel free to share your idea here)
- improve the pitch point distribution strategy. 
- Unit test

oxygen-dioxide:
Now Ramer–Douglas–Peucker algorithm is used to reduce points like how [utaformatix](https://github.com/sdercolin/utaformatix3) does. Pitchlines are splited at the position with the minimal delta to the tone of the note, which will enhance user experience, because the rapid change of pitch between notes won't be splited into two notes.

image
(The pitch here is generated by nnsvs)

oxygen-dioxide:
There is still an issue: the user have to press ctrl-Z twice to fully redo this action. One for applying the pitchpoints and one for erasing PITD.

stakira:
> There is still an issue: the user have to press ctrl-Z twice to fully redo this action. One for applying the pitchpoints and one for erasing PITD.

Seems an easy fix?

PR #669, opened 2023-04-26
Refactor PluginRunner to OpenUtau.Core and testable

arkfinn:
- Plugin logic move from ViewModel to OpenUtau.Core.
- Create interface IPlugin.( for unit test)
- Create unit test.

This fix might makes easier for plugins test in future.
But I know we have various concept on coding.
If more good thoughts, would you give me.

### DIFF
--- a/OpenUtau/ViewModels/PianoRollViewModel.cs
+++ b/OpenUtau/ViewModels/PianoRollViewModel.cs
@@ -6,6 +6,7 @@ using System.Reactive;
 using System.Security.Cryptography;
 using Avalonia.Threading;
 using DynamicData.Binding;
+using OpenUtau.Classic;
 using OpenUtau.Core;
 using OpenUtau.Core.Editing;
 using OpenUtau.Core.Ustx;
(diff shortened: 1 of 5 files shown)


OpenUtau/ViewModels/PianoRollViewModel.cs
```diff
@@ -112,36 +113,27 @@ public class PianoRollViewModel : ViewModelBase, ICmdSubscriber {
...
+                UNote? last;
+                if (NotesViewModel.Selection.IsEmpty) {
+                    first = part.notes.First();
+                    last = part.notes.Last();
+                } else {
+                    first = NotesViewModel.Selection.FirstOrDefault();
+                    last = NotesViewModel.Selection.LastOrDefault();
+                }
+                var runner = new PluginRunner(PathManager.Inst);
+                runner.OnReplaceNote += ((sender, args) => {
```
stakira, line 127: DocManager is a Core class. Maybe you don't need OnReplaceNote and OnError events? It will avoid GC complexity around event handlers.

arkfinn: Thanks.
I try to more simple method.

This was for the following reasons
1. it is the responsibility of the ViewModel to add undo/redo to plugin process.
2. DocManager has state(undoGroup), there might be a problem with parallel test.
3. To easily mocking at test.

arkfinn: @stakira 
Sorry, I understand the risks of using EventHandler in WPF.
But using DocManager from PluginRunner makes it difficult to test.
I want to remove the use of EventHandler and keep DocManager replaceable.

How about a fix like this:
- Change OnReplaceNote from event to member variable (delete event += from ViewModel)
- Initialize methods use DocManager
- Test methods that use TestMethod

example:
```
// for ViewModel
public PluginRunner(PathManager path, DocManager doc) {
  this.replaceMethod = (path, toAdd, toRemove) => {
    doc.doSomething()
  }
  ....
}
// for Test
public PluginRunner(PathManager path, Action replaceMethod) {
  this.replaceMethod = replaceMethod
  ....
}
```

arkfinn:
I tried to fix it like this

PR #679, opened 2023-05-05
Fix around language

maiko3tattun:
- Fix text was not binding to resources
- Fixed a bug that caused the language to run in the same language as InstalledUICulture the first time it was launched, but reset to en-US when the preference window was opened

### DIFF
--- a/OpenUtau/App.axaml.cs
+++ b/OpenUtau/App.axaml.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Globalization;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
 using Avalonia;
 using Avalonia.Controls.ApplicationLifetimes;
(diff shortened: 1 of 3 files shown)


OpenUtau/App.axaml.cs
```diff
@@ -59,6 +60,11 @@ public class App : Application {
             if (resDict != null) {
                 Current.Resources.MergedDictionaries.Remove(resDict);
                 Current.Resources.MergedDictionaries.Add(resDict);
+                var langName = resDict.Source.OriginalString.Replace("/Strings/Strings.", "").Replace(".axaml", "");
```
stakira, line 63: It is not meant to set the reference automatically. It is meant to follow system language if never manually set.

stakira: OK never mind. This is not unreasonable either.

maiko3tattun: Should I not save?
If I delete line 66, Language in the preference file will remain string.Empty.

maiko3tattun: I wasn't sure if I should put it in Preferences.cs, but I couldn't write the code that references the resource dictionaries.

PR #691, opened 2023-05-14
Singer window improvements, etc

maiko3tattun:
- Fix data to be passed to legacy plugins
- Fix that when the Singer window is opened, the singer used in the track is selected
- Add oto search to singer window  

Some files seem to have been changed when opened in VisualStudio, even though nothing has been changed.
(because of the BOM?)


OpenUtau.Core/Classic/UstNote.cs
```diff
@@ -38,7 +38,8 @@ class UstNote {
                 modulation = (int)phoneme.GetExpression(project, track, Ustx.MOD).Item1;
                 flags = FlagsToString(phoneme.GetResamplerFlags(project, track));
                 if (phoneme.oto != null) {
-                    filename = phoneme.oto.DisplayFile;
+                    var relativePath = phoneme.oto.File.Replace(track.Singer.Location + "\\", "");
```
oxygen-dioxide, line 41: You are hard-coding the windows path separator "\\". Does this work on platforms other than windows?

PR #698, opened 2023-05-14
[JA CVVC] Add alternate support

lottev1991:
This time, there should be no issue with null alternates. When the alternate is null, it will instead return the mapped note without alternate.
**Example:** if ``[a あ2]`` does not exist, ``[a あ]`` will be returned.


stakira:
Please add test cases for this phonemizer so that we know new features doesn't break old ones.

PR #700, opened 2023-05-14
[Default Phonemizer] Add alternate support

lottev1991:
Similar to #698 and #699.

Since there's no note mapping, null alternates shouldn't be an issue.

### DIFF
--- a/OpenUtau.Core/DefaultPhonemizer.cs
+++ b/OpenUtau.Core/DefaultPhonemizer.cs
@@ -16,7 +16,7 @@ namespace OpenUtau.Core {
             // For this simple phonemizer, all these notes maps to a single phoneme.
             string alias = notes[0].lyric;
             var attr0 = notes[0].phonemeAttributes?.FirstOrDefault(attr => attr.index == 0) ?? default;
-            if (singer.TryGetMappedOto(notes[0].lyric, notes[0].tone + attr0.toneShift, attr0.voiceColor, out var oto)) {
+            if (singer.TryGetMappedOto(notes[0].lyric + attr0.alternate, notes[0].tone + attr0.toneShift, attr0.voiceColor, out var oto)) {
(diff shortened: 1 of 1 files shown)


OpenUtau.Core/DefaultPhonemizer.cs
```diff
@@ -16,7 +16,7 @@ public class DefaultPhonemizer : Phonemizer {
             // For this simple phonemizer, all these notes maps to a single phoneme.
             string alias = notes[0].lyric;
             var attr0 = notes[0].phonemeAttributes?.FirstOrDefault(attr => attr.index == 0) ?? default;
-            if (singer.TryGetMappedOto(notes[0].lyric, notes[0].tone + attr0.toneShift, attr0.voiceColor, out var oto)) {
+            if (singer.TryGetMappedOto(notes[0].lyric + attr0.alternate, notes[0].tone + attr0.toneShift, attr0.voiceColor, out var oto)) {
```
stakira, line 19: alternate is `int?`. I don't think this is right.

stakira: I've reverted other merged alternate-related changes.

stakira: Looks like it actually works. But type coercion comes with dangers. Please at this method to `Phonemizer.cs` and use it instead.
```
        public virtual bool TryGetMappedOto(string phoneme, int tone, string color, int? alt, out UOto oto) {
            return TryGetOto($"{phoneme}{alt?.ToString() ?? ""}", out oto);
        }
```

stakira: Also, you could submit related changes in the same PR.

lottev1991: Alright, that does make sense. I will do that next time.

PR #706, opened 2023-05-16
[Phonemizer Tests] Add alternate support

lottev1991:
I noticed that alternate support was not in the PhonemizerTest API, so I decided to add it.
I do hope this works correctly, if not, I will try to look for a solution.

Related to #698, #699, and #700.


stakira:
Thanks for the goodwill, but I don't see anything getting tested. Also, I think alt should be `int?`. Please add tests with changes.

PR #710, opened 2023-05-19
[JA various phoemizers] PhoneticHint support and Unicode countermeasures

maiko3tattun:
- Support PhoneticHint : No conversion, but suffix and color are applied
- Replace NFD(か゛) with NFC(が) when entered in Unicode environment
- JA VCV & CVVC supports altarnate


stakira:
Could you add tests to show your change at work?

OpenUtau.Plugin.Builtin/JapaneseCVVCPhonemizer.cs
```diff
@@ -118,8 +118,11 @@ public class JapaneseCVVCPhonemizer : Phonemizer {
         // can probably be cleaned up more but i have work in the morning. have fun.
         public override Result Process(Note[] notes, Note? prev, Note? next, Note? prevNeighbour, Note? nextNeighbour, Note[] prevNeighbours) {
             var note = notes[0];
-            var currentUnicode = ToUnicodeElements(note.lyric);
```
stakira, line 121: Is `ToUnicodeElements()` no longer needed? How so?

maiko3tattun:
#698 , #699 , #706 is merged.

OpenUtau.Plugin.Builtin/JapaneseCVVCPhonemizer.cs
maiko3tattun: ToUnicodeElements() was only used to count characters correctly.
However, this approach does not allow が aliases to be called from か+゛.

Using Normalize() to remove surrogate characters, will allow correct character counts while resolving differences in string formatting.

maiko3tattun:
I'm creating a test, so please wait a moment...

maiko3tattun:
ok, added tests for Unicode to the three Japanese phoenemizers.

PR #718, opened 2023-05-27
Simple enunu support

rokujyushi:
## Summary of changes

* Added the ability to load models available in SimpleENUNU.

## Why make this change

* To be able to work with SimpleENUNUServer.
* To be able to load SimpleENUNU compatible models.

## What we did.

* [x] Edited VoicebankLoader to be able to load models available in SimpleENUNU.
* [x] Enabled to read settings not written in config.yaml file.

## Changes.

image

## Scope of impact.

* Loading voice banks.
* EnunuConfig setup

## How to do it.

* Put sound data of SimpleENUNU compatible models in the Voice folder

## Issues

* SingingStyle support

## Remarks

* The following is SimpleENUNUServer.
https://github.com/rokujyushi/SimpleEnunu/releases/tag/SimpleEnunu-OpenUtau-0.1.0
* Based on SimpleENUNU distributed in the following tweet.
https://twitter.com/canon_73/status/1654293958607114240


OpenUtau.Core/Enunu/EnunuConfig.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.IO;
+using System.Security.Cryptography.X509Certificates;
```
oxygen-dioxide, line 3: You may have include System.Security.Cryptography.X509Certificates by mistake. visual studio often includes this when coding even if we don't need it.

OpenUtau.Core/Enunu/EnunuConfig.cs
```diff
@@ -12,10 +13,61 @@ class EnunuConfig {
...
         }
+
+        public static RawEnunuConfig SetSimpleENUNUConfig(string location) {
+            const string tableExtension = ".table";
+            const string hedExtension = ".hed";
+            const string configYaml = "config.yaml";
+            const string modelPath = "model";
+            var config = new RawEnunuConfig();
+
+            if (Directory.Exists(Path.Combine(location, modelPath))) {
```
stakira, line 33: Code in `if` and `else` are basically the same. You could instead create an array `[location, location/model]`, then loop through it. Again, the directory existence check is unnecessary when you already check the existence of a file in it.

OpenUtau.Core/Classic/VoicebankLoader.cs
```diff
@@ -92,9 +94,22 @@ public class VoicebankLoader {
                 }
             }
             var enuconfigFile = Path.Combine(dir, kEnuconfigYaml);
+            var simpleEnunuFolder = Path.Combine(dir, kEnunuModelPath);
             if (File.Exists(enuconfigFile)) {
                 voicebank.SingerType = USingerType.Enunu;
-            } else {
+            } 
+            else if (Directory.Exists(simpleEnunuFolder)) {
+                if (File.Exists(Path.Combine(simpleEnunuFolder, kConfigYaml))) {
```
stakira, line 102: 1. The directory existence check is unnecessary when you already check file existence. Same for the below one.
2. When you have `if (...) { if (...) { voicebank.SingerType = ... } }`, there is a chance the first condition is true but the second is false. As a result `SingerType` will never be set. This should be avoided.

rokujyushi:
Corrected the part you pointed out. Please check it.

rokujyushi:
Could you check?
## Changes
- Remove unnecessary using
- Fixed VoicebankLoader processing
- Fixed processing of EnunuConfig

stakira:
Thanks!

PR #727, opened 2023-05-29
Fix exp selection changes when reloading part

maiko3tattun:
- Fixed that exp selected in the piano roll window is reselected when reloading the part

This is a somewhat forceful solution. Perhaps there is a better way.


stakira:
A proper fix is tricky. Not bad if this works.

PR #730, opened 2023-06-03
Changed the behavior of DrawPitchTool

maiko3tattun:
- Hold Ctrl to select, hold Alt to smoothen
- Mouseover does not become a hand when using DrawPitchTool and EraserTool
- Fixed related translations


OpenUtau/Strings/Strings.axaml
```diff
@@ -225,7 +225,8 @@ Warning: this option removes custom presets.</system:String>
   <system:String x:Key="pianoroll.tool.drawpitch">Draw Pitch Tool (4)
     Left click to draw
     Right click to reset
-Hold Ctrl to smoothen</system:String>
+    Hold Ctrl to select
+    Hold Alt to smoothen</system:String>
```
stakira, line 229: When updating a string, delete it from all languages so that translators will know it needs to be translated again. Otherwise, the wrong texts will not be updated.

maiko3tattun: Fixed translations.

PR #733, opened 2023-06-04
[JA CVVC] Read CV append voice color

lottev1991:
This function also exists in the Presamp phonemizer.

Sometimes, voicebanks might have voice colors that are in plain CV rather than CVVC (e.g whisper/unvoiced, death growl/screamo). With this function, it can read the CV voice color without defaulting to the base CVVC voice color.

**TODO:** Port function to VCV phonemizer (future PR).


lottev1991:
This function didn't work as expected. My apologies.

PR #738, opened 2023-06-05
Solo and mute improvements

maiko3tattun:
- Assign shortcuts to solo and mute (shift + S/M) in Main and Pianoroll windows
- Restore mute when solo is turned off (2-level mute)
- Delete unnecessary codes

Option isn't including to not cancel solo of other tracks when solo with Shift held down.
https://discord.com/channels/551606189386104834/1114932292599677010

### DIFF
--- a/OpenUtau.Core/Ustx/UTrack.cs
+++ b/OpenUtau.Core/Ustx/UTrack.cs
@@ -88,7 +88,7 @@ namespace OpenUtau.Core.Ustx {
         [YamlIgnore] public string SingerName => Singer != null ? Singer.DisplayName : "[No Singer]";
         [YamlIgnore] public int TrackNo { set; get; }
         public string TrackName { get; set; } = "New Track";
-        public bool ApparentMute { set; get; }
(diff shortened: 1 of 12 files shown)


OpenUtau.Core/Ustx/UTrack.cs
```diff
@@ -89,8 +88,9 @@ public class UTrack {
         [YamlIgnore] public string SingerName => Singer != null ? Singer.DisplayName : "[No Singer]";
         [YamlIgnore] public int TrackNo { set; get; }
         public string TrackName { get; set; } = "New Track";
-        public bool Mute { set; get; }
-        public bool Solo { set; get; }
+        public bool ApparentMute { set; get; }
```
stakira, line 91: please annotate this field with `[YamlIgnore]` to be not saved in ustx. And I think `Muted` will be a better name.

maiko3tattun: ok, I fixed it.

PR #739, opened 2023-06-06
[Breaking change to simpleEnunu] Refactor singer type detection

oxygen-dioxide:
In OpenUtau.Core/Classic/VoicebankLoader.cs, openutau currently detects voicebank types (classic, enunu, simpleEnunu or my diffsinger) by checking the existance of engine config files, like enuconfig.yaml. This "passive detection" strategy may mess up when more voicebank types are supported.
* A new key, `singer_type` is added into character.yaml to let voicebanks declare their type. It can be `Enunu` or `Classic`.
* SimpleEnunu is removed from singer type detection, as its voicebank structure need to be regulated.
* SimpleEnunu voicebanks (and any upcoming newer singer type) should declare their voicebank types in character.yaml.

I am making this change since there aren't too many enunu voicebanks distributed now, and this change won't cause a broad incompatibility. SimpleEnunu voicebank developers should add a line to their character.yaml:
```
singer_type: Enunu
```

### DIFF
--- a/OpenUtau.Core/Classic/VoicebankConfig.cs
+++ b/OpenUtau.Core/Classic/VoicebankConfig.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using OpenUtau.Core;
+using OpenUtau.Core.Ustx;
 using YamlDotNet.Serialization;
 
 namespace OpenUtau.Classic {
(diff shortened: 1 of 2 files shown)


rokujyushi:
I tried the latest beta.
singer_type: if Enunu is not described,
It is loaded in a legacy way. So SimpleENUNU voicebanks are loaded as USingerType.Classic.
It would be better to prevent SimpleENUNU voicebanks from being loaded.
Also, the constant on line 31 is unused.
https://github.com/stakira/OpenUtau/blob/a227d56f91ee663c1a61ec4b290b222b2846b2dc/OpenUtau.Core/Classic/VoicebankLoader.cs#L28-L32

PR #742, opened 2023-06-13
Display phonemes in a SingersList

rokujyushi:
## Summary of Changes

* Changed EnunuSinger.cs to set phoneme information in .hed files to Otos
* Changed USinger.cs so that it can be read even if oto.ini file and .wav file do not exist.

## Why make this change

* To be able to see the available phonemes in the ENUNU voicebank in SingersDialog.

## What we did.

* Modified USinger.cs UOtoSet class to check string.IsNullOrEmpty so that it can be instantiated even if the File of OtoSet class in Voicebank.cs is empty or null.
* Added string.IsNullOrEmpty check to USinger.cs UOto class so that it can be instantiated even if Wav of Voicebank.cs Oto class is empty or null.


rokujyushi:
Close because it contained unnecessary changes

PR #757, opened 2023-06-24
[JA Presamp/CVVC] Fix voice color fallback bug

lottev1991:
There was a bug in both of these phonemizers where, if an append voice color contains the same suffix as on the main voice color, the voice color would fall back to having no suffix on that particular pitch. Now it will fall back on the main voice color instead, as intended.


lottev1991:
I'm still working this out, it's giving me some issues in regards to CV colors.

lottev1991:
See #806.

PR #758, opened 2023-06-25
Let users choose which file types to remember in "Open Recent"

oxygen-dioxide:
Before this change, openutau remembers all the .ust and .vsqx files imported in "Open Recent". However, they are not the native file type of OpenUtau. After I imported them, I always immediately save them to .ustx and never open the original .ust or .vsqx again in OpenUtau. So here I add an option to disable remembering them in "Open Recent".

The default behaviour is unchanged: Remembering .ust and .vsqx but not remembering .mid

image

### DIFF
--- a/OpenUtau.Core/Util/Preferences.cs
+++ b/OpenUtau.Core/Util/Preferences.cs
@@ -38,6 +38,33 @@ namespace OpenUtau.Core.Util {
             Save();
         }
 
+        public static void AddRecentFileIfEnabled(string filePath){
+            //Users can choose adding .ust, .vsqx and .mid files to recent files or not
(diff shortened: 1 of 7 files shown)


OpenUtau.Core/Util/Preferences.cs
```diff
@@ -38,6 +38,34 @@ public static class Preferences {
...
+                    if(Preferences.Default.RememberVsqx){
+                        AddRecentFile(filePath);
+                    }
+                    break;
+                default:
+                    break;
+            }
+        }
+
         public static void AddRecentFile(string filePath) {
```
stakira, line 69: Please make this private so that it won't get used accidentally.

PR #762, opened 2023-07-01
Fix solo and mute

maiko3tattun:
New features:
- Solo additionally (which not removes solo from other tracks)
- Unsolo all
- Mute this only
- Mute all others
- Unmute all

UI fix:
- Freeze volume and pan when track is muted
- Mute button works independently of solo

Specification:
- When a track is mute and solo, priority is given to the solo


maiko3tattun:
Please correct sentences if English is wrong!

PR #772, opened 2023-07-10
For fine control of volume and panning.

rokujyushi:
## Summary of Changes

* Adjustment of track header layout
* Allow input of volume and pan values

## Why make this change

* Because the slider is difficult to operate.
* To be able to make fine adjustments.

## What we did.

* Add controls to TrackHeader.axaml
* Added show, hide, and input handling to TrackHeader.axaml.cs

## What you are changing
image

## Challenges

* Layout is not very clear.
* Program implementation is simple, but lacks maintainability.
* Might be inefficient.

Translated with www.DeepL.com/Translator (free version)


rokujyushi:
I think we need to address that, because even if you mute it, you can still operate it.

stakira:
1. It's hard to see what you changed. Could you revert unrelated changes?
2. Can the original TextBlock handle double-click to show the TextBox?
3. I believe you can check object equality instead of name equality.
4. Pleas try to style the TextBox to the same font size as the TextBlock.

rokujyushi:
TextBlock is to be avoided if possible, since the area of the click will be the size of the text.
Other than that, I understand.

PR #778, opened 2023-07-15
Fix voice color resetting

maiko3tattun:
In the UExpression class, minimum and maximum values of voice color are no longer checked.
This prevents the colors to reset when duplicating tracks or using legacy plugins.


stakira:
Are you sure this doesn't crash? What happens when CLR is out of range?

maiko3tattun:
I did some tests, such as changing to voice bank with fewer voice colors.
It does not crash, but none of the colors are selected.

stakira:
How would it render then? No suitable oto?

PR #785, opened 2023-07-17
Add Japanese contracted sound processing to EditLyrics

maiko3tattun:
Bug fix : Add Japanese contracted sound processing in SplitLyrics class

Before
image

After
image


stakira:
I don't understand what this does, but I'll trust a native Japanese speaker on this.

PR #792, opened 2023-07-28
Fix Note properties(RE) and Add Vibrato drift

maiko3tattun:
New Features:
- Textboxes and sliders now work
- Vibrato drift (move height)
This is a parameter that has existed internally in ustx up to now, but has not been used.
Currently the only way to change this parameter is to manipulate the note property.
image

Not yet implemented:

- Right click on the slider does not work
(PointerPressed does not work well on sliders, so Focus is used instead, and it cannot determine if it is a right click or not.)

### DIFF
--- a/OpenUtau.Core/Commands/NoteCommands.cs
+++ b/OpenUtau.Core/Commands/NoteCommands.cs
@@ -310,6 +310,30 @@ namespace OpenUtau.Core {
         }
     }
 
+    public class VibratoDriftCommand : VibratoCommand {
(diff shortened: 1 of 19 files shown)


maiko3tattun:
To fix the look of expanders, the style of the toggle button has been modified.

stakira:
conflict....

maiko3tattun:
Should I merge or rebase upstream to this? Or should I create a new branch and PR?

PR #805, opened 2023-08-20
Update it-IT translation

IDOLTRASH:


IDOLTRASH:
Idk but somethings wrong from appveyor with the translation. I'll try again

PR #810, opened 2023-08-26
Add Lengthen crossfades in batch menu

maiko3tattun:
- New Feature: Lengthen crossfades
Lengthen the overlaps regardless of the original overlap of the oto.
Also lengthen the preutter if it can be lengthened as a result.
image
image
image

- Why is this feature needed?:
The oto overlap for OG UTAUs is sometimes intentionally set short(1/3 of preutter).
This is to avoid going crazy when the previous note is short.
However, in actual singing, if it can be lengthened, it will sound better.
For this reason, plugins have been developed since ancient times to lengthen crossfades, and the function has been included in 連続音一括設定(most famous plugin for VCV conversion) and presamp.

### DIFF
--- a/OpenUtau.Core/Editing/NoteBatchEdits.cs
+++ b/OpenUtau.Core/Editing/NoteBatchEdits.cs
@@ -282,6 +282,51 @@ namespace OpenUtau.Core.Editing {
         }
     }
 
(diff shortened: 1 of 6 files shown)


stakira:
Great stuff!

PR #824, opened 2023-09-06
[JA Presamp Phonemizer] Fix glottal stop VCV & accidental "empty" VC when none is present

lottev1991:
- Glottal stop VCV featuring ``・`` will now properly connect to the next note (both ``[v ・V]`` and ``[v V・]``);
- Fix accidental "empty" VC insertion when none is present in the bank.

**TODO:** Fix glottal stop CV?

PR #827, opened 2023-09-09
[JA Presamp Phonemizer] Supports "っ", etc

maiko3tattun:
New features:
- Add replaces ぢ, づ, with じ, ず
- Support for "っ"

Improvement:
- When there is a phonetic hint, the lyrics are not replaced


maiko3tattun:
The "っ" consonant lengthening symbol.
If the "a っ" is present, it is converted to it; if not, it is automatically converted to VC.
image

maiko3tattun:
Added an improved version of #824 .

PR #833, opened 2023-09-10
Add option to set/change singer avatar

lottev1991:
This function was discussed and requested, so I decided to add it.

You can now set/change the singer avatar from the Singers menu. Note that it will not change on the track header until the singer gets manually reloaded.
Option to set singer avatar


lottev1991:
Duplicate, closing

PR #834, opened 2023-09-11
Check and Remapping voice color

maiko3tattun:
New Features:
- Voice color remapping dialog
image
image
- Voice color remapping menu in track header context menu
- When the project is opened, show the remapping dialog if the voice color lineup has been changed
- When changing voice banks, show the remapping dialog if the voice color lineup is different

Internal change:
Stores the name of the voice colors in the UTrack class


stakira:
wow this is cool. does it also handle out of range values?

maiko3tattun:
Yes, out of range phonemes of older mappings fall back to default values.

PR #838, opened 2023-09-12
Correction of structure names, etc.

rokujyushi:
## Summary of Changes

* Changed the name of the structure
* Change the value used in conditional branches

## Why make this change

* Because the original name was inappropriate for the function.
* To change the processing on the SimpleENUNUServer side as well.


stakira:
@oxygen-dioxide could you please help review?

PR #840, opened 2023-09-14
Refactor search dialog to a searchbar like in browsers and text editors

oxygen-dioxide:
- Refactor search dialog to a searchbar, like in browsers and text editors
- Add Key bindings: Enter for navigating to the next match, Shift+Enter for navigating to the previous match, and ESC to exit searchbar.
- If the user selected a note before opening the search dialog, use the lyric of the selection as the search word.
- Show both the count of matches and the index of current match.

image


oxygen-dioxide:
TODO: 
- Full word matching
- search within phonemes
- Case insensitive mode

PR #852, opened 2023-09-20
[Singers window] Add button to show readme.txt

lottev1991:
There's always been a way to show the character.txt/character.yaml, but there was never a way to directly view the readme file. Since the readme often contains important information (such as a ToS, usage tips, etc.), I've decided to add a button to open this file.

To limit clutter on the Singers window, the readme file will open in the default external application of the operating system. If a readme.txt does not exist, a popup window will appear with the text "readme.txt not found".

readme.txt button

readme.txt not found dialog

Showing readme.txt in external program

### DIFF
--- a/OpenUtau/Views/SingersDialog.axaml.cs
+++ b/OpenUtau/Views/SingersDialog.axaml.cs
@@ -195,7 +195,7 @@ namespace OpenUtau.App.Views {
         void OnOpenReadme(object sender, RoutedEventArgs e) {
             var viewModel = (DataContext as SingersViewModel)!;
             if (viewModel.Singer != null) {
-                var readme = $@"{viewModel.Singer.Location}\readme.txt";
+                var readme = Path.Join(viewModel.Singer.Location, "readme.txt");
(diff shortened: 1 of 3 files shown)


OpenUtau/Views/SingersDialog.axaml.cs
```diff
@@ -191,6 +192,27 @@ public partial class SingersDialog : Window, ICmdSubscriber {
             }
         }
 
+        void OnOpenReadme(object sender, RoutedEventArgs e) {
+            var viewModel = (DataContext as SingersViewModel)!;
+            if (viewModel.Singer != null) {
+                var readme = $@"{viewModel.Singer.Location}\readme.txt";
```
oxygen-dioxide, line 198: Backslash only works on Windows. use `var readme = Path.Join(viewModel.Singer.Location,"readme.txt")` so that it is cross-platform.

lottev1991: Thanks. Fixed.

PR #853, opened 2023-09-21
[Singer window] Play sample on click + [Wave.cs] Support reading AIFF format

lottev1991:
This adds a button to play the sample file of a voicebank (doesn't have to be a wav file, as long as it's a supported audio format). If a sample is not configured, it will play a random file from the voicebank folder (including subfolders). Upon clicking the button again and/or when the Singer window is closed, the sample will stop playing. It will also stop playing upon switching the singer and when opening the Voice Color menu.

This PR also adds support for reading AIFF audio files; I had coded this extension into the sample player, only to find out that OpenUtau did not support it (which is why the two are combined in one PR). Some known UTAU-compatible voicebanks (such as the old Macloids) are known to use AIFF as their audio format of choice; since it's a common format on Mac (and is supported on other platforms as well), I decided it was a good idea to make OpenUtau support it. This will aid cross-platform support. (Only a few resamplers support AIFF, but this is up to the end user.)

[Short demonstration video with sound](https://drive.google.com/file/d/1zAVLMtrKWzCa6vqa3Zu92aILoHF7KkuA/view?usp=sharing)

[Randomized audio file demonstration](https://drive.google.com/file/d/1ttYhCcJpUoH11JBz3FWMcOcvLNoMLJrI/view?usp=sharing)

### DIFF
--- a/OpenUtau.Core/Classic/ClassicSinger.cs
+++ b/OpenUtau.Core/Classic/ClassicSinger.cs
@@ -28,6 +28,7 @@ namespace OpenUtau.Classic {
         public override float PortraitOpacity => voicebank.PortraitOpacity;
         public override int PortraitHeight => voicebank.PortraitHeight;
         public override string DefaultPhonemizer => voicebank.DefaultPhonemizer;
+        public override string Sample => voicebank.Sample == null ? null : Path.Combine(Location, voicebank.Sample);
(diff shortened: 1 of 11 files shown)


lottev1991:
Forget about it.

lottev1991:
Reopening just for testing. I doubt it will succeed. It will remain a draft for the time being

PR #858, opened 2023-09-27
Fix voicebank error check

maiko3tattun:
Improvements:
- When loading voicebanks, setting an error and "oto.IsValid = false" for aliases if wav does not exist.
- Improved singer's error report:
sample rate, channels, and alias duplicates are now reported.
image

### DIFF
--- a/OpenUtau.Core/Classic/VoicebankErrorChecker.cs
+++ b/OpenUtau.Core/Classic/VoicebankErrorChecker.cs
@@ -60,7 +60,7 @@ namespace OpenUtau.Classic {
                 });
             }
             try {
-                VoicebankLoader.LoadVoicebank(voicebank, false);
+                VoicebankLoader.LoadVoicebank(voicebank);
             } catch (Exception e) {
                 Errors.Add(new VoicebankError() {
                     message = "Failed to load voicebank",
--- a/OpenUtau.Core/Classic/VoicebankErrorChecker.cs
+++ b/OpenUtau.Core/Classic/VoicebankErrorChecker.cs
@@ -60,7 +60,7 @@ namespace OpenUtau.Classic {
                 });
             }
             try {
-                VoicebankLoader.LoadVoicebank(voicebank);
+                VoicebankLoader.LoadVoicebank(voicebank, false);
             } catch (Exception e) {
                 Errors.Add(new VoicebankError() {
                     message = "Failed to load voicebank",
--- a/OpenUtau.Core/Classic/VoicebankErrorChecker.cs
+++ b/OpenUtau.Core/Classic/VoicebankErrorChecker.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
+using NAudio.Wave;
 
 namespace OpenUtau.Classic {
     public class VoicebankError {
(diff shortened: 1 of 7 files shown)


maiko3tattun:
I'll have to modify the tests...

maiko3tattun:
I had to write dirty code to pass the tests ; ;

OpenUtau.Core/Classic/VoicebankLoader.cs
```diff
@@ -317,7 +319,7 @@ public class VoicebankLoader {
                             otoSet.Otos.Add(oto);
                         }
                         if (!string.IsNullOrEmpty(oto.Error)) {
-                            Log.Error($"Failed to parse\n{trace}: {oto.Error}");
+                            Log.Error($"Failed to parse\n{oto.Error}");
```
stakira, line 322: Why remove this? The whole point of `trace` is to help people see where the errors are from.

OpenUtau.Core/Classic/VoicebankLoader.cs
```diff
@@ -386,6 +388,13 @@ public class VoicebankLoader {
                 oto.Error = $"{trace}\nFailed to parse overlap. Format is {format}.";
                 return oto;
             }
+            if (!IsTest) {
+                string path = Path.Combine(Path.GetDirectoryName(trace.file), oto.Wav);
+                if (!File.Exists(path)) {
```
stakira, line 393: It's wasteful to check for every oto line. It would be better to parse all lines first, then check files.

OpenUtau.Core/Classic/VoicebankErrorChecker.cs
```diff
@@ -108,6 +117,29 @@ public class VoicebankErrorChecker {
             try {
                 using (var wav = Core.Format.Wave.OpenFile(filePath)) {
                     fileDuration = wav.TotalTime.TotalMilliseconds;
+                    var waveFormat = wav.ToSampleProvider().WaveFormat;
```
stakira, line 120: Alternative, conversion can be done like how non-wav files are handled. See https://github.com/stakira/OpenUtau/blob/05076e487213dca32660c378e83299b22de17182/OpenUtau.Core/Classic/VoicebankFiles.cs#L83-L93

OpenUtau.Core/Classic/VoicebankLoader.cs
maiko3tattun: That's because `trace` is included in `oto.Error`, so it's a double.

OpenUtau.Core/Classic/VoicebankErrorChecker.cs
maiko3tattun: This is an error report and should be reported down to the smallest detail.
In fact, the current worldline will not play correctly if the sample rate is incorrect. There is also a possibility that voicebanks created by OU could be distributed and used by OG UTAU.

OpenUtau.Core/Classic/VoicebankLoader.cs
maiko3tattun: I updated it.

maiko3tattun:
Minor errors in error reports are separated as information.
This is because information such as missing character.yaml is misinterpreted as if it were a serious error.

Currently, stereo files that are supported by worldline are reported as information, not 44100 Hz files that are not supported by worldline are reported as errors.
It is true that worldline should support a variety of sample rates, but that is a discussion for another place.

image

PR #862, opened 2023-10-02
Fixed classic pitch

delta-kimigatame:
Improvements:

* Add:Base64 testing.
* Changed:run-length compression  to produce results similar to pure UTAU.
* Changed Classic pitch algorithm :nearboarhood to linear interpolate

image


stakira:
Delta さん saving the world.

PR #864, opened 2023-10-03
Changed sample playback button to be displayed only in Classic

rokujyushi:
https://github.com/stakira/OpenUtau/pull/863
I saw the above modification and thought it should be changed to not display if it is only used in classic in the first place.

MinWidth was changed because when the window is displayed at its smallest size, the controls cover it.

### DIFF
--- a/OpenUtau/Views/SingersDialog.axaml
+++ b/OpenUtau/Views/SingersDialog.axaml
@@ -50,7 +50,7 @@
         <DataGridTextColumn Header="{StaticResource singers.subbanks.color}" Binding="{Binding Color}"/>
       </DataGrid.Columns>
     </DataGrid>
-    <StackPanel Grid.Row="0" Spacing="10" Margin="10,0,0,0" Height="20"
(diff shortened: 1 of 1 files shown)


Astel123457:
this doesn't make sense? because 1. WORLDLINE-R still uses classic banks, 2. ENUNU/NNSVS voicebanks can contain samples, (eg ritsu kire ai)

lottev1991:
Not to mention that ENUNU banks usually have a character.txt file. These work the exact same as for a classic UTAU singer, including support for configuring an audio sample.

#864 provides a better solution for this problem.

rokujyushi:
OpenUtau cannot synthesize ENUNU voicebangs with UTAU voicebangs included.
I think the UTAU voice bang included with ENUNU was to support UST creation.
Therefore, it is not necessary for OU.
Many of the currently distributed voicebanks include the Namine Ritsu voicebangs.
If the function is to listen to voicebank samples, it should play the specified wav, not the UTAU voicebang.

image
image

PR #874, opened 2023-10-07
Scale degree display

oxygen-dioxide:
Users can select a key for each ustx project, such as 1=D.

image

Then they can see scale degrees of each tone on their piano roll. There are two display styles:
1. Solfege (do re mi fa sol la ti)
image

2. Numbered (1 2 3 4 5 6 7)
image


stakira:
Is there an "off" status?

oxygen-dioxide:
> Is there an "off" status?

Yes

https://github.com/stakira/OpenUtau/blob/8cddfbb012062ad197df2d7df39365308cda65af/OpenUtau/Views/PreferencesDialog.axaml#L156-L160

oxygen-dioxide:
It is off by default. Users who want this feature have to turn it on manually in preferences.

PR #880, opened 2023-10-09
(WIP DON'T MERGE) VCCV Phonemizer Vowels Update

AnAndroNerd:
Adds support for [9l] [8n]. This doesn't work very well yet. If you can, please help me. I have no clue what I'm doing.

KNOWN ISSUES

- The phonemizer doesn't spilt into CCs correctly.

Bounce becomes [b8n] [8ns], it should be [b8] [8n-] [ns-]

- The phonemizer doesn't do VV transitions correctly.

What it does is it makes the VVs [8n n] [nx], which should be [8n-] [nx]. [9l] VV transitions don't work correctly at all for [9l] because what should be [9l l] is [9 l l] in VCCV English OTOs, and using [9 l l] for transitions is still wrong, and it should be [9l-]


AnAndroNerd:
I removed the Vowel [Ang] it's a pain to get working correctly.

PR #887, opened 2023-10-14
Check wav file writeable when exporting

maiko3tattun:
Check if wav cannot be written when exporting tracks or mixdown because they are in use or for other reasons.


stakira:
I don't see a reason to not interrupt the render and show the error to the user. The text on the progress bar won't last more than 1 second. Hiding the error leads to problematic render without the user knowing what's happening.

maiko3tattun:
Show error, and stopped using task.Wait().

stakira:
What's the goal of this PR? Is it to avoid "not writeable" errors or make them less annoying? Are such errors transient (e.g. another thread is writing the file) or persistent? I don't think the error should be simply dismissed, and I don't quite see how changing the way to show the error helps.

Only if the error is transient, waiting a bit and try write that file again seems a meaningful change. Otherwise rendering need to stop and the user needs to be informed clearly.

maiko3tattun:
In the current OU, if a file fails to be written, the message "Exporting to path~. is displayed and nothing happens.
When rendering fails, also, there is no error message, just a log.
If the user inadvertently opens a wav, or if the render fails, we will not notice that the file was not updated.

PR #889, opened 2023-10-16
[WIP: Do not merge] Prefer VCV when using same subbank

adlez27:
EN to JA phonemizer
- Prefer VCV when there is 1 consonant and prev/next note use the same subbank

Syllable based phonemizer
- Add phoneme attributes to Syllable and Ending
- Add overloaded method of `AreTonesFromTheSameSubbank` which accepts voice color


adlez27:
image

known issues
- checks for exact same pitch instead of same pitch range within voice color

adlez27:
I will not be adding this feature yet. To work reliably it requires many changes outside of the EN to JA phonemizer and it will be difficult to review. Those other changes will come first.

PR #892, opened 2023-10-20
Show loading dialog / Set try-catch

maiko3tattun:
Improvements:
- Show dialog when heavy processing
- Set try-catch for some processes to avoid crashes


stakira:
This is a nice change, but I don't think it should be done with a notification. As you have found, I'd prefer a message box modal created and destroyed locally like in RegenFrq().

maiko3tattun:
This may be called from the ViewModel, so they cannot find the parent window...

PR #898, opened 2023-10-22
[KO VCV Phonemizer] Voice color fix + alternate support + extend end breath support

lottev1991:
- Fixes voice color and tone shift by mapping the aliases. Legacy mapping has been removed, as it is no longer necessary;
- Add alternate support;
- Extend end breath support (``R2``). Also works through alternate.


Blackmaster1988 (changes requested):
- [ ]

PR #906, opened 2023-10-24
Fix TextBox font

maiko3tattun:
The font used for the text boxes is a bit hard to read in Japanese, so changed to the same font as the text and labels.
It may depend on the environment, but it will not have any adverse effect.
This is largely a personal preference, so please merge if necessary.
image

### DIFF
--- a/OpenUtau/Styles/Styles.axaml
+++ b/OpenUtau/Styles/Styles.axaml
@@ -62,7 +62,6 @@
     <Setter Property="Padding" Value="4,1"/>
     <Setter Property="Margin" Value="0,0,0,4"/>
     <Setter Property="FontSize" Value="12"/>
-    <Setter Property="FontFamily" Value="{DynamicResource ui.fontfamily}"/>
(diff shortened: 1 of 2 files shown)


OpenUtau/Styles/Styles.axaml
```diff
@@ -62,7 +62,7 @@
     <Setter Property="Padding" Value="4,1"/>
     <Setter Property="Margin" Value="0,0,0,4"/>
     <Setter Property="FontSize" Value="12"/>
-    <Setter Property="FontFamily" Value="{DynamicResource ui.textbox.fontfamily}"/>
+    <Setter Property="FontFamily" Value="{DynamicResource ui.fontfamily}"/>
```
stakira, line 65: These was for working around Avalonia 10 could not handle CJK font fallback in text boxes. You can simply delete this line and delete `ui.textbox.fontfamily` from `Strings.axaml`

maiko3tattun: Ok! updated.

PR #907, opened 2023-10-25
Fix note property panel

maiko3tattun:
Improvements:
- Sliders can be reset by right clicking
- Update text box when lost focus instead of immediately
- "WORK IN PROGRESS" has been removed as the required functions are now available.

Bug fix:
- Reflects portamento and vibrato presets edited by note defaults
- Reflects changes in project expression


maiko3tattun:
Tone can also be controlled here!

PR #908, opened 2023-10-25
Add "ArpaXG2p", "DiffSinger English X Phonemizer"

spicytigermeat:
- I've added a new G2p model, "ArpaXG2p", which is an English G2p Model that supports the [ax] and [dx] phoneme.
  - More refining for the model will come in time :)
- I've added support to the PhoneticHelper and Lyric Tool to use "ArpaXG2p"
- Added "DiffSinger English X Phonemizer", which is a phonemizer that
  - Utilizes "dsdict-enx.yaml" in DS Voicebank folder like other DS Phonemizers
  - Uses "ArpaXG2p" for phonetic estimation of words outside of the dictionary


hetteric:
(Opinion)
It's great to see a phonemizer that supports the full phoneme set, however it may be best to also provide a more universal g2p and not only one specifically designed for how your single Diffsinger model is labeled.

For example, your model doing words like "tree" as [ch r iy] instead of [t r iy] is nonstandard and will cause incompatibility with nearly all other English Diffsinger models.

Astel123457:
This is because of how the g2p was trained, not because of how the model was labeled. The dictionary that the g2p was based off of uses these "bad" pronunciations and doesn't contain alternates like the original. As far as I know, they don't have one and just used an open source dictionary.

hetteric:
> This is because of how the g2p was trained, not because of how the model was labeled. The dictionary that the g2p was based off of uses these "bad" pronunciations and doesn't contain alternates like the original. As far as I know, they don't have one and just used an open source dictionary.

For clarity, I was referencing what was said in an online chat server. In which the dictionary (if I recall correctly, a subset of CMU dict) was specifically edited and trained for their personal acoustic model.
The differences in pronunciation are as such because they had made the personal choice to handle the phonetics uniquely from other existing acoustic models.

spicytigermeat:
> For clarity, I was referencing what was explained in an online chat server. In which a subset of CMU dict was specifically modified and trained for their personal acoustic model. The differences in pronunciation are as such because they had made the personal decision to handle the phonetics uniquely from other existing acoustic models. So this G2P model is specific to them and their labeling style alone as the preexisting one wouldn't fit.

While I understand your concern (as we've discussed), I find it a bit of a far reach to say that the way my G2p handles one cluster of phonetics to be cause for "universal incompatibility" to be a very strange and unfriendly sentiment that has to do with your personal preference. The G2p is trained on a custom edited version of the CMU dictionary where there **is** consistency in how the onset cluster in ```Tree [ch r iy]``` is handled. All instances of this phoneme are ```[ch r]``` except for proceeding a ```[s]``` where it stays as ```[s t r]``` instead, as general American pronunciation follows that according to my observation. I do not think this would cause any inconsistencies because any complete and working English voice library will have ```[t]``` and ```[ch]```. The manner you've presented your concern is also sad to see as OpenUTAU is a collaborative tool, and your comments have made this contribution feel unwelcome.

hetteric:
> While I understand your concern (as we've discussed), I find it a bit of a far reach to say that the way my G2p handles one cluster of phonetics to be cause for "universal incompatibility" to be a very strange and unfriendly sentiment that has to do with your personal preference. The G2p is trained on a custom edited version of the CMU dictionary where there **is** consistency in how the onset cluster in ```Tree [ch r iy]``` is handled. All instances of this phoneme are ```[ch r]``` except for proceeding a ```[s]``` where it stays as ```[s t r]``` instead, as general American pronunciation follows that according to my observation. I'd also like to add that the G2p was not trained specifically for my personal model, but any model labelled with ```[ax]``` and ```[dx]```, I'm not sure where that comment is coming from. I do not think this would cause any inconsistencies because any complete and working English voice library will have ```[t]``` and ```[ch]```. The manner you've presented your concern is also sad to see as OpenUTAU is a collaborative tool, and your comments have made this contribution feel unwelcome.

Your wording in the OU server very much came across as it existing only for your model. Given that you've said things such as "`i personally use [ch r] for [tr] and [jh r] for [dr] but that's my accent`" and other remarks about changing the phonemes per-accent.

If that's not the case then I apologize.

I'm _not_ saying it shouldn't be included, however my concern is with the majority of preexisting models that do not do this.
I think if this were to be included then a second G2P that handles the vast majority of models that do [d r] and [t r] also needs to be included.

spicytigermeat:
> Your wording in the OU server very much came across as it existing only for your model. Given that you've said things such as "`i personally use [ch r] for [tr] and [jh r] for [dr] but that's my accent`" and other remarks about changing the (other) phonemes per-accent.

Understandable, but there was a lot of other conversation going on and I think maybe my comment was taken out of context. That is how I label according to my accent, but many other users labeling English agreed with that but also some agreed with `[t r]`, so it is hugely up to preference. Either way, whether a user is using a voice labeled for `[ch r]` or `[t r]`, they're going to 1: Get audible results and 2: both options will sound similar.

Taking this to assume my G2p is only for my voice libraries, however, is inherently incorrect as if I was to create a G2p for my labeling process, I would include things like `[tcl]`, `[q]` for glottal stops and other plosive modifications (`[k cl]` and `[p cl]`) but decided that to ensure compatibility across other voices, I am choosing to omit them and instead specify them in my `dsdict.yaml` for each phonemizer I'm deciding to support. Clarification was never explicitly asked for so I apologize if my replies came off as confusing.

> I'm _not_ saying it shouldn't be included, however my primary concern is with the majority of preexisting models that do not do this. I think if this were to be included then a second G2P that handles the vast majority of models that do [d r] and [t r] also needs to be included. My secondary concern is with wanting to change the phonemes dialectically, which would mean this particular G2P may see limited usability.

The distributors of other models are well within their rights to set the default phonemizer on their preference, and can provide recommendations for users, but ultimately its the users decision how they use the voice libraries. Either way, the appearance of `[t r]` in English while not uncommon is not going to make or break a track if one voice works best with `[ch]` and another works with `[t]`, as OpenUTAU is not at the point where vocals sound on the same level as commercial products with little to no editing. Therefore, the point made here seems to be more of a personal opinion rather than concern for the general user-base of OpenUTAU. Regardless, you're training your own G2p so all of the issues you've brought up will be completely null once your G2p is implemented, as the biggest difference from the samples you showed in the DiffSinger server was that your dictionary was further along in development and uses `[t r]` and `[d r]` for the plosive + liquid cluster. So, the point you were making felt more of a personal issue directly from you rather than anything else.
 
> At no point did I insult, belittle, or say your g2p should not be included. From the beginning I only suggested the inclusion of a second G2P. However, by providing your phonemizer to potentially add to OpenUTAU, an open source project with many users, it is understood that it may be up for discussion or concern with how it may affect existing users.

I agree completely! Your reception was cold, however, so I apologize for taking your comments out of context, and wish you luck on your own G2p!

stakira:
oops, could you help resolve the conflict?

PR #911, opened 2023-10-29
Audio transcription with SOME

oxygen-dioxide:
This PR adds integration with [SOME](https://github.com/openvpi/some), an audio to musical notes converter. Users can right-click on a wave part and select "Transcribe" to convert it into a voice part.
image

image

Before using, please download the SOME model from https://github.com/xunmengshe/OpenUtau/releases/download/0.0.0.0/some-0.0.1.oudep


wolfgitpr:
@maiko3tattun May I ask if the loading dialog can be used here.

maiko3tattun:
> @maiko3tattun May I ask if the loading dialog can be used here.

Perhaps it could be used.
However, MessageBox.ShowModal() might be better than LoadingNotification(#892). It is used to create FRQs.

yqzhishen:
@oxygen-dioxide Maybe we should add checks to prevent very long chunks after slicing, for example, if someone clicks "transcribe" on an accompaniment audio that has no silence parts at all. The model contains attention operators and can cause OOM on very long sequences.

OpenUtau/Views/MainWindow.axaml.cs
```diff
@@ -1057,6 +1061,40 @@ public partial class MainWindow : Window, ICmdSubscriber {
...
+                    using(Some some = new Some()){
+                        string text = ThemeManager.GetString("context.part.transcribing");
+                        var msgbox = MessageBox.ShowModal(this, $"{text} {part.name}", text);
+                        //Duration of the wave file in seconds
+                        int wavDurS = (int)(wavePart.fileDurationMs / 1000.0);
+                        var transcribeTask = some.Transcribe(DocManager.Inst.Project, wavePart, wavPosS =>{
+                            //msgbox?.SetText($"{text} {part.name}\n{wavPosS}/{wavDurS}");
+                            msgbox.SetText(string.Format("{0} {1}\n{2} / {3}",text, part.name, wavPosS, wavDurS));
+                        });
+                        transcribeTask.Wait();
```
stakira, line 1076: Instead of `Wait()` use `ContinueWith()`. See `RegenFrq()` for example. Modify project and show/hide dialog on scheduler from `TaskScheduler.FromCurrentSynchronizationContext()`.

OpenUtau/Strings/Strings.axaml
```diff
@@ -8,6 +8,8 @@
   <system:String x:Key="context.part.delete">Delete part</system:String>
   <system:String x:Key="context.part.rename">Rename part</system:String>
   <system:String x:Key="context.part.replaceaudio">Reselect audio file</system:String>
+  <system:String x:Key="context.part.transcribe">Transcribe</system:String>
```
stakira, line 11: Even though `transcribe` is correct here, a more specific description like "Transcribe audio to create a note part" could be helpful.

PR #912, opened 2023-10-29
Add "View" menu in PianoRollWindow

maiko3tattun:
- Add "View" menu in PianoRollWindow
Some preference menus can be changed on the piano roll window.
image
- Update each controls when settings are changed


stakira:
Conflict

PR #922, opened 2023-11-15
Use filename as alias

maiko3tattun:
New Features:
- Use voice bank filenames as aliases
This is false by default.
This feature is very important for some CV voicebanks.
image
image

- Added a menu in the SingerWindow to set whether or not to use this feature
This can be set for each voicebank and will be written in the character.yaml.
image


stakira:
conflict

PR #924, opened 2023-11-16
[EN X-SAMPA Phonemizer] Consonant length optimization + refactor

lottev1991:
I have adjusted ``GetTransitionBasicLengthMs()`` in the phonemizer by defining the oto suffix (both pitch and voice color) as nullable. This way, it gets ignored when calculating the transition length through ``if (alias.EndsWith(c))``.


lottev1991:
I am reopening this with a better branch. Apologies.

PR #929, opened 2023-11-18
Set SingerType in SingersWindow

maiko3tattun:
Write the singer type in character.yaml.
This PR should be merged after #892.
image


maiko3tattun:
Pressing this menu with the Vogen singer will result in an error, which will be fixed on another occasion along with other errors.

PR #897, opened 2023-10-22
add mandarin/cantonese g2p

wolfgitpr:
Improvements:
1、Removed the original Tinypinyin, and the new ZhG2p supports Mandarin and Cantonese.
2、Using the public song dataset opencpop for annotation (approximately 194 thousand words) testing, the error rate decreased from 2.33% of the original (TinyPinyin) to 0.06%.
TinyPinyin：
image
New G2p:
image

PR #934, opened 2023-11-20
[BaseChinesePhonemizer] Use PinyinHelper instead of WordsHelper for hanzi conversion

lottev1991:
For the record, I am aware of #897. However, I couldn't help but notice that the Base Chinese Phonemizer API uses WordsHelper rather than TinyPinyin's PinyinHelper (which the PinyinLyricsHelper uses). PinyinHelper tends to give the correct pinyin results, whereas WordsHelper gives plenty of errors. (For example, the common particle 么 gets romanized as "mo", when it is actually "me". PinyinHelper does not have this issue.)


lottev1991:
I'll just close this, it's no longer needed.

PR #938, opened 2023-11-22
[WIP] Add Singer Selection Dialog

maiko3tattun:
### Overview:
This is a huge modification.
For some time now, users who own many voice banks have had difficulty managing them.
This PR will improve the following two points: 
- Allow to set multiple paths for singers, and make it easier to manage directories.
- Display a dialog to select a singer and allow users to filter and search singers.

https://discord.com/channels/551606189386104834/1131900332092436611

### New Features:
- Singer Selection Dialog
image
  - This appears instead of the context menu when selecting a track's singer.
  - Categorize singers by recently used, favorites, and directories.
Users can easily pin their favorites by pressing the heart icon next to their name.
Recently used singers are sorted by date, all others by localized name.
  - Categorize singers by type, such as classic, enunu, etc. These can be expanded.
  - Easily reload singers.
  - Allows setting the range of folders to be used.
  Previously, all directories within singer's locations were searched, but now there is an option to search only top directories as in OG UTAU.
- Singer Location Dialog
image
  - Allows users to set up additional singer paths and locations where new singers are to be installed.
  - This can be called from Singer Selection Dialog and Preferences.

### Internal Changes:
- From now on, AdditionalSingerPath"s" will be used instead of AdditionalSingerPath in preferences.
- Singers with the same folder name are considered the same. If the same folder exists in different locations, only one of them will be loaded.
This follows the previous specification that treated folder names as IDs, but this remains controversial.
Technically, USinger implements IEquatable.
- Favorites are saved in preferences as well as recently used ones.

### DIFF
--- a/OpenUtau.Core/Classic/ClassicSingerLoader.cs
+++ b/OpenUtau.Core/Classic/ClassicSingerLoader.cs
@@ -17,11 +17,7 @@ namespace OpenUtau.Classic {
         }
         public static IEnumerable<USinger> FindAllSingers() {
             List<USinger> singers = new List<USinger>();
-            foreach (var path in new string[] {
-                PathManager.Inst.SingersPathOld,
-                PathManager.Inst.SingersPath,
-                PathManager.Inst.AdditionalSingersPath,
-            }) {
+            foreach (var path in PathManager.Inst.SingersPaths) {
                 var loader = new VoicebankLoader(path);
                 singers.AddRange(loader.SearchAll()
                     .Select(AdjustSingerType));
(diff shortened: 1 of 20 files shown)


maiko3tattun:
We can choose to select the singer in the conventional context menu.
Also, we can select a singer in the dialog with one click.

yqzhishen:
Can the singer selection context menu be preserved to show recently used (or "pinned") voicebanks, and meanwhile add something like a `More...` button to open this dialog for detailed selections?

nagotown:
I would still like to see this as an option.

PR #946, opened 2023-11-25
Fix exception catching in audio transcribe

maiko3tattun:
Bug fix:
- Fixed a bug that failed to catch exceptions that occurred during audio transcribe task.
- Able to determine if the exception is due to SOME uninstallation.


maiko3tattun:
image

PR #948, opened 2023-11-26
Fix creation of empty Oto from blank line in otoini

maiko3tattun:
Fixed a bug that a blank line in oto.ini created an empty oto, which caused voice banks to fail to load.


stakira:
please update the test if it's expected

PR #960, opened 2023-11-30
[Chinese CVVC Phonemizer] Add support for Cantonese (hanzi conversion)

lottev1991:
With this implementation, the phonemizer checks for specific aliases in the oto before romanizing hanzi characters. Depending on this, it will convert to either pinyin or jyutping.

I might add Cantonese support to the Chinese CVV phonemizer as well, but it's quite geared towards Mandarin right now and would only function properly for Cantonese with a custom dictionary. This is due to aliasing conflicts, especially regarding tail phonemes. The Chinese CVVC phonemizer only needs a presamp.ini file.


lottev1991:
I got some feedback that it's a better idea to make a separate Cantonese phonemizer, so I'm closing this.

Issue #970, opened 2023-12-05
Phonemizer stopped Auto-inputing VCs


Michelino98:
### Acknowledgement

- [X] I have read Getting-Started and FAQ

### 🐛 Describe the bug

Hello like the title says. OU's Phonemizer stopped automatically input the VCs or even on other banks when they're used as voicecolor. Only if the CVVC VB is stand alone it normally uses it's own VCs multi or monopitch. I remember for example if you bundle and voice color assing'em, normally.
It should be:
 - あ扇風機 a s扇風機 さ扇風機 a k楽 こ楽 i ん扇風機 n ny悲 に悲 i ch楽 ち楽 i w楽 わ楽 
and not as:
 - あ扇風機 a s さ扇風機 a k こ楽 i ん扇風機 n ny に悲 i ch ち楽 i w わ楽.

I notices that when was making a bundle VB for (空音ライ)Sorane Lie using they're other banks made by 三ヶ木ちい(Mikage chii) and color assing each bank.
I wanted to test out they're CVVC append named 贋作 bank, that's the moment I saw for the first time the issue. 
Here some screenshots.
![phoneticnazers are ignoring CVVC's VCs 2](https://github.com/stakira/OpenUtau/assets/137897357/5ee8419f-0faa-4780-846b-54e55de8f370)
***Not only it happens with CVVC*** but with VCVLiteCVVC too (they're rare btw) like Honoka Elie's 女子アナウ(Girl Annoucer)the one with 宣 Suffix. The VCs are needed for the full use.
![phoneticnazers are ignoring CVVC's VCs 1 ver 2](https://github.com/stakira/OpenUtau/assets/137897357/2f41205c-e228-48db-b182-779d7ead6077)
the next example is about not automaticlly inputing the correct VC when it's fully CVVC supported including appends. it's what I wrote earlier.
![phoneticnazers are ignoring CVVC's VCs 3](https://github.com/stakira/OpenUtau/assets/137897357/8e109b37-83f5-4722-ba3b-e029b25ec92b)
not even the EN to JA is safe from the issue.
![Even the Engrish](https://github.com/stakira/OpenUtau/assets/137897357/86d0e45d-fe02-407b-98e4-20adea12daf1)

[...]

PR #975, opened 2023-12-09
[Japanese Presamp Phonemizer] Add option for MUSTVC

lottev1991:
This makes it so that when ``[MUSTVC]`` is set to ``1`` in the presamp.ini file, a VC phoneme will be inserted even on short notes. When set to ``0`` (which is the default), this will not happen.

### DIFF
--- a/OpenUtau.Plugin.Builtin/JapanesePresampPhonemizer.cs
+++ b/OpenUtau.Plugin.Builtin/JapanesePresampPhonemizer.cs
@@ -195,7 +195,7 @@ namespace OpenUtau.Plugin.Builtin {
             // Insert 2nd phoneme (when next doesn't have hint)
             if (nextNeighbour != null && string.IsNullOrEmpty(nextNeighbour.Value.phoneticHint)) {
                 int totalDuration = notes.Sum(n => n.duration);
-                if (TickToMs(totalDuration) < 100) {
+                if (TickToMs(totalDuration) < 100 && presamp.MustVC == false) {
                     return MakeSimpleResult(currentLyric);
                 }
 


Michelino98:
@lottev1991 Would this be able to fix the missing VCs from phonemizers? glad it was taken care but how can I do it? a tutorial would be much apprecieted. (^ ^ )

lottev1991:
> @lottev1991 Would this be able to fix the missing VCs from phonemizers? glad it was taken care but how can I do it? a tutorial would be much apprecieted. (^ ^ )

I don't know what you mean with "phonemizers" as this PR only applies to one specific phonemizer. This function never existed in any of the other phonemizers in the first place.

Michelino98:
Oh I Apologize, I'm refering the Phonemizer in plural since in OU their are called like: [JA CVVC] Japanese CVVC Phonemizer, [EN VCCV] English Phonemizer, along with the contributor in paretesis and so on. Felt like their multiple Phonemizer. to the Issue that I'm refering here's the link: https://github.com/stakira/OpenUtau/issues/970
I might made into a mess of examplanation but it shows my probleam. And gonna be honest I mainly use mostly the JA VCV&CVVC Presamp Phonemizer hehe.

lottev1991:
> Oh I Apologize, I'm refering the Phonemizer in plural since in OU their are called like: [JA CVVC] Japanese CVVC Phonemizer, [EN VCCV] English Phonemizer, along with the contributor in paretesis and so on. Felt like their multiple Phonemizer. to the Issue that I'm refering here's the link: #970 I might made into a mess of examplanation but it shows my probleam. And gonna be honest I mainly use mostly the JA VCV&CVVC Presamp Phonemizer hehe and also I was flabbergasted by seeing the appends's VC not showing.

The issue you've linked to seems to be unrelated to this PR.

PR #976, opened 2023-12-09
Fix singer window goes behind the main window

maiko3tattun:
Fixed singer window being shown in the back when opened from the main window menu.


maiko3tattun:
Additionally, fixed behavior when activating the Singer window from the Piano Roll window.
The track's singer is displayed even when no alias is selected.

PR #989, opened 2023-12-26
Add KoreanPhonemizerUtil

EX3exp:
- Added KoreanPhonemizerUtil.cs, and This class will be used in **Diffsinger Korean Phonemizer**, **other Multiple Korean Phonemizers(including Classic)**.
- Implementing Korean Phoneme Variation(like 꽃망울에 to [꼰망우레]), Jamo separation(like 간 to ㄱㅏㄴ), Jamo merge(like ㄱㅏㄴ to 간) in Each phonemizers makes Korean Plugin developers struggle.... So I made KoreanPhonemizerUtil, to increase Productivity and Code re-using.
- Also supports .ini management, currently used by KO CV Phonemizer & KO ENUNU Phonemizer.

- If this branch merged, I want to PR two Classic Korean Phonemizers and Diffsinger Korean Phonemizer which using this class. 👀 (+ [23.12.28] Korean ENUNU Phonemizer uses this class too)

### DIFF
--- a/OpenUtau.Core/KoreanPhonemizerUtil.cs
+++ b/OpenUtau.Core/KoreanPhonemizerUtil.cs
@@ -287,6 +287,10 @@ namespace OpenUtau.Core {
                     firstLastConsonant = "ㅂ";
                     nextFirstConsonant = "ㅅ";
                 } 
+                else if (firstLastConsonant.Equals("ㄻ")) {
+                    firstLastConsonant = "ㄹ";
+                    nextFirstConsonant = "ㅁ";
+                } 
                 else if (firstLastConsonant.Equals("ㅇ") && nextFirstConsonant.Equals("ㅇ")) {
                     // Do nothing
                 } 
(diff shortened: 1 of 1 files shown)


oxygen-dioxide:
Why do you add a new ini parser implementation instead of using the existing code in openutau https://github.com/stakira/OpenUtau/blob/master/OpenUtau.Core/Classic/Ini.cs ?

oxygen-dioxide:
Also if you're adding custom dictionary support, yaml is a better format because it is more extensible, and most other phonemizers use yaml.

EX3exp:
> Why do you add a new ini parser implementation instead of using the existing code in openutau https://github.com/stakira/OpenUtau/blob/master/OpenUtau.Core/Classic/Ini.cs ?

Oh... It was my mistake...😲 I'll fix ini parsing implementation right away. Thanks for mentioning it!

EX3exp:
> Also if you're adding custom dictionary support, yaml is a better format because it is more extensible, and most other phonemizers use yaml.

I either think .yaml format would be better choice... I'll change dictionary's format into yaml. Thanks for your advice!

EX3exp:
I Fixed BaseIniManager's implementation to use OpenUtau's `Ini.cs`, and now KO ENUNU Phonemizer uses user-customed Jamo dictionary in Yaml format(`jamo_dict.yaml`). 👀

PR #992, opened 2023-12-26
Add KoreanCVPhonemizer

EX3exp:
- Added Korean CV Phonemizer(CLASSIC).
- Since Korean CV Voicebanks have many variations in Alias, It uses `Ko-CV.ini` to Handle Various CV voicebanks.
(for example, 시 is used as "shi" in some VBs and "si" in others, Batchim 아-ㄴ is used as "an" in some VBs and "a n" in others... so on. We are calling ALL of them as "Korean CV Voicebanks", but They are commonly using Similar, but Different phoneme symbols in some part.) 
- Supports Phoneme hints input(like 냥[nya, a ng]), Detailed phoneme position adjustments(Especially at end consonants) for better synthesize quality, and Korean Phoneme Variation.
- This is Branch after (Should be merged after) #991

### DIFF
--- a/OpenUtau.Core/KoreanPhonemizerUtil.cs
+++ b/OpenUtau.Core/KoreanPhonemizerUtil.cs
@@ -1,0 +1,1433 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using OpenUtau.Core.Ustx;
+using OpenUtau.Classic;
+using Serilog;
+using static OpenUtau.Api.Phonemizer;
+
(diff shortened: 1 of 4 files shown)


EX3exp:
I just found a number of bugs now, so turned all korean phonemizer PRs into draft.... 🫠 Soon I'll fix codes and simplify algorithms.

OpenUtau.Core/KoreanPhonemizerUtil.cs
stakira, line 1: You could move this class to the plugin project.

stakira: never mind

PR #999, opened 2024-01-01
Create winget.yml

Exorcism0666:
All you have to do is create a "WINGET_TOKEN" token with the necessary scopes, and bingo! you've just automated the output of your software for winget. :)

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -1,0 +1,13 @@
+name: Publish to WinGet
+on:
+  release:
+    types: [released]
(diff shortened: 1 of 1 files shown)


oxygen-dioxide:
There are two winget package names for openutau: "stakira.OpenUTAU" and "stakira.OpenUTAU.Beta". The former one is for stable versions and the latter one is for beta versions. so you need to detect whether the release is a pre-release to determine which package to update.

Exorcism0666:
normally it's good!

PR #1005, opened 2024-01-06
Properties panel: Numeric calculation and click bug fix

maiko3tattun:
New Feature:
- Move values relative to each other
Allows positive or negative numbers to be entered in text boxes.
movie: https://github.com/stakira/OpenUtau/assets/130257355/9691d85a-b503-4fa5-8751-4ed3caffdc46

Bug fix:
- Fixed behavior of single click on slider


stakira:
Does the movie show using "+50" to add 50 to all values? What about negative values? How to determine if the user want to subtract or fill a negative value?

PR #1006, opened 2024-01-10
DiffSinger: support different dictionaries for Chinese, Japanese and Cantonese

oxygen-dioxide:
- DiffSinger voicebanks can use different dictionaries for Chinese (dsdict-zh.yaml), Japanese (dsdict-ja.yaml) and Cantonese (dsdict-zh-yue.yaml). For example, DiffSinger Chinese phonemizer tries to load dsdict-zh.yaml first. If dsdict-zh.yaml isn't found, it will load dsdict.yaml as before. This will make it easier to develop multilingual voicebanks.
  - This change doesn't affect existing voicebanks.
- Add DiffSinger Japanese phonemizer (DIFFS JA). Currently the only difference between DIFFS JA and DIFFS is that DIFFS JA tries to load dsdict-ja.yaml before loading dsdict.yaml.
- Moved diffsinger phonemizer code to a subfolder.


stakira:
conflict

PR #1008, opened 2024-01-10
Add UnixCaseCheck in voice bank check

sdercolin:
Implement UnixCaseCheck to check duplicates like `a.wav` and `A.wav` which would not work on Windows.


stakira:
Could you unify the check despite OS to ensure:
1. Case matches.
2. No files that only differ in case.
The check doesn't block normal uses. So let's simply be strict, so that checked VB works everywhere.

sdercolin:
@stakira Please check again!
cc @oxygen-dioxide

PR #1013, opened 2024-01-17
Add Millfeuille G2P and Fr DiffSinger phonemizer

imsupposedto:
Add Millfeuille G2P trained by myself
Implementation help by tigermeat
Ensured G2P is in lyric helper and phonetic assistant

### DIFF
--- a/OpenUtau/Views/MessageBox.axaml
+++ b/OpenUtau/Views/MessageBox.axaml
@@ -8,59 +8,61 @@
         Title="MessageBox" MinWidth="300" MinHeight="150"
         WindowStartupLocation="CenterScreen"
         ExtendClientAreaToDecorationsHint="False">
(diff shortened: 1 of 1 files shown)


OpenUtau.Core/Util/LyricsHelper.cs
```diff
@@ -1,112 +1,117 @@
-﻿using System;
```
stakira, line 1: Could you please try to make the line-ending styles match? Should be CRLF.

imsupposedto: I'm just about done with finals at uni so I'll take a look soon. I'm not very proficient in code so it may take me some time if I need to find someone to help.

lottev1991:
Is this still being worked on? I think it would be great to have this phonemizer built-in, due to the popularity of Millefeuille banks.

PR #1034, opened 2024-02-06
Support "- C" in JA VCV&CVVC Phonemizer

maiko3tattun:
This change does not affect other phonemizers.
~~It may not be compatible with existing ustx that use this Phonemizer.~~

Summary:
- Phonemizer may or may not specify a phoneme index
- ~~It is possible that there is no phoneme with index 0~~
- It can be used like this: [- s][さ]


maiko3tattun:
This can be merged as I have adjusted it so that it does not affect the existing ustx.

PR #1041, opened 2024-02-16
Open voicebank icon file read-only

liuycsd:
Open voicebank icon file read-only.

Fix an "[ERR] Failed to load avatar data" issue if current user can not open the file for writing.


oxygen-dioxide:
Why not also make this change for utau and enunu voicebanks?

PR #1043, opened 2024-02-17
Add a feature to only render and play selected region

liuycsd:
This pull request tries to add a feature that only renders and plays selected region in the PianoRollWindow.

Only a simple play-region command via Alt+Space is provided in this PR.
It seems to work. But the UI interaction may need further improvement.

### DIFF
--- a/OpenUtau.Core/PlaybackManager.cs
+++ b/OpenUtau.Core/PlaybackManager.cs
@@ -81,21 +81,25 @@ namespace OpenUtau.Core {
             return sineGen;
         }
 
-        public void PlayOrPause() {
+        public void PlayOrPause(int tick = -1, int endTick = -1, int trackNo = -1) {
             if (Playing) {
                 PausePlayback();
             } else {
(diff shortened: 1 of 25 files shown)


stakira:
Great feature!

However, I'm a little doubtful about soloing to a single track with any note selection. This makes users unable to listen to a section from all tracks. Maybe you don't need to add it. If a user wants to solo, they still have the option in the tracks view.

liuycsd:
> Great feature!
> 
> However, I'm a little doubtful about soloing to a single track with any note selection. This makes users unable to listen to a section from all tracks. Maybe you don't need to add it. If a user wants to solo, they still have the option in the tracks view.

OK. Fixed.

One more commit is added to skip rendering muted tracks (instead of calculating all the notes on those tracks then scaling the volume to zero). Although it's not perfect because it's only checked once when it starts rendering so canceling mute on those already skipped tracks won't actually work before current playback stops.
So I add an option for this behavior.

Besides, would it be better to use something like start / end markers on the time axis to mark the region to play, for all tracks?

PR #1048, opened 2024-02-24
Add "no value" to expression

maiko3tattun:
In implementing track expression and phonemizer expression in the future, we need to clearly distinguish whether a phoneme has a value or not.

### Changes:
- To distinguish between "set value to 100" and "no value (default value of 100)", SetNote/PhonemeExpressionCommand will accept null as well as float.
- In the expression panel and properties panel, the presence or absence of a value can be visually determined.
image
- Implement logic for "phonemizerExpressions".

### What I want to implement in the future:
The ability for the phonemizer to recommend an expression(voicecolor, P flag, etc.) and for the user to override it.
https://github.com/stakira/OpenUtau/assets/130257355/85e0e071-bc48-4417-9661-0827158eb5e7
(prototype)

### DIFF
--- a/OpenUtau.Core/Commands/ExpCommands.cs
+++ b/OpenUtau.Core/Commands/ExpCommands.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-
 using OpenUtau.Core.Ustx;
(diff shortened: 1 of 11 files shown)


stakira:
Cool stuff!

PR #1052, opened 2024-02-26
Add phonemizer expression

maiko3tattun:
New Feature:
- Phonemizer expression:
Phonemizer can suggest expressions such as flags and volumes. Expressions specified in the UI will always take precedence over those suggested by the phonemizer.
If phonemizer suggest an expression that is not in the project, it will be ignored.

Changes:
- Phonemizers can optionally get projects and tracks: Phonemizer.SetUp()
It is possible to get voice colors, and to get the track expression when it is implemented in the future.
- JA VCV&CVVC Phonemizer
The color of VC will be that of CV unless otherwise specified. In the future I would like to adjust the volume of consonants.

Fix:
- Fixed a bug in the Paste Selected Parameters menu where it would only paste to the first phoneme when the note to paste to had no expression: UNote.cs int[] phonemeIndexes
This was probably there before.

### DIFF
--- a/OpenUtau.Core/Api/Phonemizer.cs
+++ b/OpenUtau.Core/Api/Phonemizer.cs
@@ -97,10 +97,21 @@ namespace OpenUtau.Api {
             public string voiceColor;
         }
 
+        public struct PhonemeExpression {
+            public string abbr;
+            public float value;
+        }
+
         /// <summary>
         /// The output struct that represents a phoneme.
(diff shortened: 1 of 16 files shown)


maiko3tattun:
Fix JA VCV&CVVC Phonemizer:
- Fixed bug that unwanted main VC appears when using CV voice color
- Fixed a case where the previous lyric is "っ"

PR #1063, opened 2024-03-12
Add "Lock editing of unselected notes" menu

maiko3tattun:
Please merge after #1058.

New Feature:
- Added a menu option to allow or disallow editing of notes that are not selected
image

Background:
- Expressions originally allowed editing only on selected notes, but there was a need to edit regardless of selection.
- There was a problem of confusion when the pitch curve to be edited overlapped with the pitch curve or vibrato of other notes.

Please let me know if there is something wrong with the English in the menu or code.


stakira:
conflict

PR #1069, opened 2024-03-19
Fix Draw Pitch Tool

maiko3tattun:
The pitch curve is gotten from RenderPhrase, not the note, so that PITD can be drawn correctly with draw pitch tool even if there is vibrato.
image

Merging with #1068 makes it more useful!


stakira:
That was on purpose. If you don't want vibrato, turn it off. On the other hand, you may want to change the overall shape while keeping the vibrato.

maiko3tattun:
It is disconcerting to have the pitch line drawn away from the mouse cursor.
It is also useful to be able to use the Draw Pitch tool to draw an overwrite when we want to modify a portion of the vibrato.

stakira:
Could you make it an "alternative tool" like pen-plus?

PR #1086, opened 2024-03-29
Translate error messages and put them in the expander

maiko3tattun:
image
Expander is closed by default.

### DIFF
--- a/OpenUtau/Views/MessageBox.axaml
+++ b/OpenUtau/Views/MessageBox.axaml
@@ -13,7 +13,11 @@
       <RowDefinition Height="*"/>
       <RowDefinition Height="50"/>
     </Grid.RowDefinitions>
+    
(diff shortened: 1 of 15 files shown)


OpenUtau/Views/MessageBox.axaml
```diff
@@ -39,8 +43,15 @@
             </ControlTemplate>
           </Setter>
         </Style>
+        <Style Selector="Expander">
+          <Setter Property="Margin" Value="0, 20, 0, 0" />
+          <Setter Property="Width" Value="560" />
+          <Setter Property="IsExpanded" Value="False" />
```
stakira, line 49: I'd prefer it expanded by default. The stack trace was added so that users' screenshots are actually useful. If they screenshot without expanding it then you will have to ask them to do so, etc, etc, which wastes energy.

PR #1089, opened 2024-03-30
Add overwrite option to DrawPitchTool

maiko3tattun:
This is an improved version of #1069.
- Add "Draw Pitch Tool to overwrite vibrato or MOD+" option to edit menu
Pitch curves with vibrato or MOD+ applied can be modified with the Draw Pitch tool.

https://github.com/stakira/OpenUtau/assets/130257355/0b6acfba-353a-4117-9d2f-9ae147e5a9ef


stakira:
Let's not add a tool option at a weird place... you can add a button next to the draw pitch tool button, like how pen plus is next to pen. You can use the same icon as draw pitch tool for now.

PR #1099, opened 2024-04-08
[Feedback wanted] DiffSingerKoreanPhonemizer overhaul/code trimming + add dedicated ``dsdict-ko.yaml`` support

lottev1991:
Basically, I decided to simplify this phonemizer down to the bare minimum where the configurations of lyrics and phonemes will be defined by a dedicated dictionary rather than be hardcoded in the phonemizer. This is because different DiffSinger banks use different phonetic systems, and might want a dedicated dictionary which does not conflict with any other dictionaries the bank may have.

This has 3 purposes:
- It will get a dedicated ``dsdict-ko.yaml`` file;
- It gives more freedom to the end user to define their own dictionary entry and phonemes, instead of predefining what are vowel-, semivowel- or batchim phonemes;
- It trims down the code significantly, leaving only the bare minimum necessary.

Alternatively, I am willing to simply add support for ``dsdict-ko.yaml`` and leave the rest of the code as-is, but due to its sheer length and the hardcoded features I have to admit it does not have my personal preference. However, as I'm not the original developer of this phonemizer, I don't want to do anything that developers and users may not want.


stakira:
@EX3exp any feedback?

EX3exp:
@lottev1991 Oh, I'm so Sorry for late reply... I agree with your comment almost all,  and I was about to fix rest of this phonemizer's code.(but couldn't because of my midterm😂... It ends on this wednesday)

**I think this could make you confusing, but this phonemizer is not actually using any hardcoden phonemes...! It uses grapheme-phoneme dictionary from singer's `dsdict.yaml` when generating phonemes** (like my screenshot below)
image

**All of variables and functions including predefined phonemes(like *IsPlainVowel*) will be deprecated, and they were used when forcing phoneme's timing, rather than aligning phonemes**! (*I think forcing phoneme's timing was most dumb features that I've done... I'll delete them soon, so please ignore them...*)

**So If user wants to apply one's own phonetic system, all they have to do is modifying the `dsdict.yaml` correctly**. 
[Recently I could apply different Kor phonetic system, without using any additional dictionaries.](https://github.com/EX3exp/ex3-ko-diffsinger-support/blob/main/dict/Readme.md)
**It seems no need to make additional dictionaries, I think dictionaries are enough with `dsdict.yaml`...**

When I commited this phonemizer, I was have to implement Korean Phoneme Variation(음운변동) features, So I had to modify Korean Lyrics before it was passed to singer's g2p(that made from `dsdict.yaml`). So I Copy&Pasted some line from  [DiffsingerBasePhonemizer.cs](https://github.com/EX3exp/OpenUtau/blob/master/OpenUtau.Core/DiffSinger/DiffSingerBasePhonemizer.cs) and used as base code, 
And the result is these long heavy code...(God I'm sorry) I think this caused more misunderstandings...🫠

**Feature that this phonemizer needs is simply one: conducting Phoneme Variation**.
I think Phoneme Variation could be some confusing concept to other language users... please understand, we need this feature... 
- **The simple description of Phoneme Variation feature is like this**:
> a. *user input lyric*: 말, 했, 었, 던
> b. *changed lyric* after phoneme variation: 마, 래, 써, 떤 
> c. Then *changed lyric* will be passed to Singer's g2p(generated from dsdict.yaml, and this is already implemented in DiffSingerBasePhonemizer.), then phonemizer aligns phonemes with singer's own phonetic system.

**My feedback is**:
- **Could you trim down codes (like you did) without adding `dsdict-ko.yaml` features?**
Then this phonemizer will be cannot conduct Korean Phoneme Variation for a while, but currently this phonemizer has one annoying bug: **generates 'error' to whole phrase when non-Korean Letter exists**... and I think this could cause some uncomfortance to users... 
Current codes needs to be refactored, also It needs some different approaches for Phoneme Varition Implementation. (I'm thinking to clean up and rewrite whole code, with more simplified logics.)

**I think I'll add new Phoneme Variation implementation soon, but If you're willing to implement that, please try!
Since Implementing Phoneme Variation to each phonemizer is such a pain for programmers, I previously made Phoneme Variation Functions in [KoreanPhonemizerUtil.cs](https://github.com/stakira/OpenUtau/blob/master/OpenUtau.Core/KoreanPhonemizerUtil.cs#L185)... (Please consider using that)**

lottev1991:
I've restored the old code for now, aside from the custom dsdict which I kept.

Perhaps some classes in the base phonemizer could be made public in order to trim down the code here. But I'm not going to do that if it's not desired.

lottev1991:
No longer needed, so I'm closing it.

PR #1100, opened 2024-04-08
Fix deleting expressions

maiko3tattun:
- When ustx is loaded, expressions that have been deleted from the project are also deleted from the note (because a null expression descriptor in the note causes an error): UNote
- Some expressions that was getting directly from the project are now getting via track (Supplement to #1017)
- ~~When expressions are deleted in ExpressionsDialog, the default required expressions are added again: ProjectCommand
Pending as I don't have a solution for duplicate flags.~~


maiko3tattun:
A solution was made in #1118 for when a required expression is deleted in the Expressions dialog.

PR #1101, opened 2024-04-09
[EN X-SAMPA] Vowel fallback bug fixes

lottev1991:
- It has come to my attention that I accidentally removed a Teto-specific vowel fallback some time ago (where any instance of ``V`` when combined with ``b`` will fall back to ``@``). This PR fixes this and re-implements the fallback.
- Fixed a split diphthong bug.
- Fixed issue where VV's weren't used in diphthongs.


lottev1991:
I'm currently fixing those; I did not add that specific code, but I just noticed this today.

PR #1109, opened 2024-04-22
[WIP don't merge] Load custom plugins from singer folders

lottev1991:
Adds the ability to load plugins from the singer folders, e.g. custom phonemizers. This might make it easier to use plugins specifically tailored to one specific voicebank.

I'm not entirely sure if this function will fit within the scope of the software; there has certainly been demand, but I'm okay with it if it's not desired.

**TODO:** Avoid loading duplicate files? (Still researching)

### DIFF
--- a/OpenUtau.Core/DocManager.cs
+++ b/OpenUtau.Core/DocManager.cs
@@ -77,6 +77,9 @@ namespace OpenUtau.Core {
                     File.Delete(oldBuiltin);
                 }
                 files.AddRange(Directory.EnumerateFiles(PathManager.Inst.PluginsPath, "*.dll", SearchOption.AllDirectories));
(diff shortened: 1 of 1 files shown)


lottev1991:
I've been trying to get it to load plugins from the folder of the currently loaded singer, however this does not seem to be working. I've tried any combination of `SetSinger()`, `singer.Location` and`singer.BasePath` (I admittedly don't know the difference). None of these combinations seem to work. Any ideas? Do I need to define something else? Is there a way to return the folder name of the currently loaded singer, or is that what the aforementioned classes are for? I'm a little bit confused. If it won't work out I'm afraid I'm gonna have to close this PR, unless I can find a solution.

PR #1124, opened 2024-05-01
Fix LyricsReplaceDialog bug

maiko3tattun:
Fixed preview failure when lyrics replacement contains characters that cannot be used in regular expressions.


stakira:
Conflict

PR #1125, opened 2024-05-01
Fixed force activation of the main window when an error dialog is shown

maiko3tattun:
- MessageBox.ShowError() always parented to the main window, but now parented to the currently active window.
- Reduce code by using optional arguments.


stakira:
What is this for?

maiko3tattun:
Since all ErrorMessageNotifications are picked up by MainWindow and the notification does not contain information about which window should be the parent, this was the quickest way to determine this.

PR #1130, opened 2024-05-08
Fix importing project into an unsaved project

maiko3tattun:
Fixed an failure to get the relative path from an unsaved project to a wave part when trying to track import with wave part into an unsaved project. Ignore this exception if already got the relative path from the original project.


stakira:
It does not make much sense to use the same relative path for the imported project and the unsaved project?

maiko3tattun:
It is inevitable that AfterLoad will occur twice for the project we want to import and for the entire project after import. Therefore, I aim to avoid errors on the second AfterLoad.

PR #1149, opened 2024-05-19
[JA VCV] Support * あ

maiko3tattun:
- VCV: The CheckOtoUntilHit used in other phonemizers is used here to increase flexibility.
- CVVC: Remove unnecessary codes.

@lottev1991 I thought this code was unnecessary as it will always be true, is this ok?
`(color ?? string.Empty) == color)`


lottev1991:
> @lottev1991 I thought this code was unnecessary as it will always be true, is this ok? `(color ?? string.Empty) == color)`

Should be fine; I believe I didn't write that part (I think it was Hitcoder?), I just optimized it later on and left that part as-is.

PR #1150, opened 2024-05-19
[JA Presamp Phonemizer] Some starting consonant fixes/adjustments

lottev1991:
This PR fixes 2 main things:
- Starting consonants will no longer be inserted before syllables that start with a priority consonant. This is because most priority consonants (usually stops and affricates) are not configured in a way to handle this in Japanese voicebanks (if the end user wishes to include stop consonants anyway, e.g. when they're configured similarly to in Arpasing banks, simply remove them from under the ``[PRIORITY]`` line in the presamp.ini);
- If a voice color does not contain a starting consonant in the oto, it will no longer be inserted (similar to how VC works). Now, it will insert a starting consonant from the main color if it exists, which may give undesired results.
- Adjust VC and starting C length based on tempo changes, similar to #1161. I'm gonna look into turning this into the default behavior (though I will keep the old method for legacy functionality).

### DIFF
--- a/OpenUtau.Plugin.Builtin/JapanesePresampPhonemizer.cs
+++ b/OpenUtau.Plugin.Builtin/JapanesePresampPhonemizer.cs
@@ -215,7 +215,9 @@ namespace OpenUtau.Plugin.Builtin {
                 && preCFlag
                 && !currentLyric.Contains(vcvpad)
                 && presamp.PhonemeList.TryGetValue(currentAlias, out PresampPhoneme phoneme)
-                && phoneme.HasConsonant) {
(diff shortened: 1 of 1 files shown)


maiko3tattun:
Thinking about starting consonant again, since there are only consonants, there is not much difference between each voice color, and it would be acceptable to fall back on the main?

lottev1991:
Not necessarily. For example, I have a voicebank with a devoiced/whisper voice color, and one with a death growl voice color. Both are special voice types and CV-only; using starting C with them would sound odd. That's why I decided on this change.

PR #1155, opened 2024-05-25
Fix OpenUtau failed to search singers when the user runs OpenUtau for the first time

oxygen-dioxide:


maiko3tattun:
In PreferencesViewModel.cs line130, singer names are
- SortingOrder == null → translated into the display language(default)
- SortingOrder == string.Empty →not translated(InvariantCulture)
- SortingOrder == each language　→translated into the specified language

maiko3tattun:
Thanks!🥰

PR #1161, opened 2024-05-29
ZH CVVC: Fix note timing after tempo change

oxygen-dioxide:
Before this fix, ZH CVVC phonemizer converts between tick and ms only based on the initial tempo instead of the changed tempo. Now it converts them based on the tempo at the position of the note


oxygen-dioxide:
For phonemizer developers: Please stop using `MsToTick` to convert durations. It should only be used on positions.

PR #1166, opened 2024-06-04
Add recovery project function

maiko3tattun:
image

In this improvement, the OU saves the recovery path in the preferences at autosave and deletes path upon successful close.
The next time it is booted, if the recovery path is still there, it is considered to be in need of recovery.


maiko3tattun:
image
dialog updated.

stakira:
Conflict...

maiko3tattun:
Delayed the timing of InitProject() because if the project is loaded before the main window is visible, the error dialog when the project fails to load and the recovery message box cannot be displayed.

PR #1173, opened 2024-06-10
Update Strings.pl-PL.axaml

KluelessKisa:
Updated the Polish translation as a large chunk of strings were simply missing a translation

hopefully nothing get accidentally broken in the strings


KluelessKisa:
What exactly is a "workflow awaiting approval" this request appears to be a bit stuck with, if I may ask?

PR #1179, opened 2024-06-13
[KoreanPhonemizerUtil] Fixed crashing when reading ini from non-utf-8 singers

EX3exp:
- `KoreanPhonemizerUtil.cs`uses singer's encoding to read .ini file, but crashing occurs by inproper key reading  when singer's encoding does not support some characters in .ini file. (e.g. When singer's encoding is Shift-jis, but singer's `ko-CV.ini` contains Hangeul characters.)
- So Forced `BaseIniManager`'s read encoding to *utf-8* to handle that issue.
- This change has effects to **Korean CV**, **Korean ENUNU** Phonemizers.

### DIFF
--- a/OpenUtau.Core/KoreanPhonemizerUtil.cs
+++ b/OpenUtau.Core/KoreanPhonemizerUtil.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text;
 using OpenUtau.Core.Ustx;
 using OpenUtau.Classic;
 using Serilog;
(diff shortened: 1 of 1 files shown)


stakira:
Wouldn't this still crash sometimes, i.e., when the ini file is indeed shift-jis?

EX3exp:
> Wouldn't this still crash sometimes, i.e., when the ini file is indeed shift-jis?

I noticed that issue just now, so fixed StreamWriter's encoding(and other StreamReader's encoding) into utf-8. Now it would handle utf-8 files only, both of reading and writing processes.

stakira:
I don't understand. Are these ini files only written and read by `KoreanPhonemizerUtil.cs`? Not something a random voicebank already has?

EX3exp:
> I don't understand. Are these ini files only written and read by `KoreanPhonemizerUtil.cs`? Not something a random voicebank already has?

Yes, those ini files are only written and read by `KoreanPhonemizerUtil.cs` (*e.g. KO CV Phonemizer's `ko-cv.ini`*), and only used in OpenUtau.
There's no Korean Voicebanks that using specific ini setting file(*e.g. `presamp.ini`  for Japanese CVVC*) in classic UTAU, as far as I know.

In case of `ko-cv.ini`, It's purpose was to cover Korean CV UTAUs' various forms of oto aliases. For readability, It is containing a few Hangeul characters in its Key value, and Hangeul characters cannot be read or written in shift-jis(or other many encodings)... 
It caused crashing issue when singer's encoding was not supporting Korean chars(*e.g. shift-jis in many cases*). (even if singer's `ko-cv.ini` itself has no encoding issues.)
And I think there's maybe few shift-jis `ko-cv.ini`, because`ko-cv.ini` could only have created by OpenUtau's KO CV Phonemizer.

PR #1186, opened 2024-06-17
Check and patch moreconfig.txt when launching

oxygen-dioxide:
Many users encountered "Unsupported file type" issue when rendering using moresampler with wrong moreconfig. After this change, OpenUtau will check moreconfig.txt and fix it when launching.

### DIFF
--- a/OpenUtau.Core/Classic/ExeResampler.cs
+++ b/OpenUtau.Core/Classic/ExeResampler.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


OpenUtau.Core/Classic/ExeResampler.cs
```diff
@@ -34,6 +36,27 @@ internal class ExeResampler : IResampler {
...
+            }
+            for (int i = 0; i < lines.Count; i++) {
+                if (lines[i].StartsWith("resampler-compatibility")) {
+                    if(lines[i] == "resampler-compatibility on"){
+                        //moreconfig.txt is correct
+                        return;
+                    } else {
+                        lines[i] = "resampler-compatibility on";
+                        File.WriteAllLines(moreConfigPath, lines);
+                        return;
```
stakira, line 52: Does this break the moresampler as both resampler and wavtool use case?

Cadlaxa: even the option `resampler-compatibility` is on, both moresampler as a wavtool and resampler still functions normally

PR #1193, opened 2024-06-23
Add dark-L vowels to EnXSampaPhonemizer.cs

HitCoder9768:
for reclists such as my own using phonemes such as `O5` for "cold"


oxygen-dioxide:
The GitHub action tests failed. Please ensure that the unit tests passed. You can use `dotnet test` command to run the tests locally.
If your code changes the behaviour of the phonemizer, also change them in the unit tests.

lottev1991:
(Apologies, I sent this on Discord too but it's kinda important) I also noticed that you've defined the `Ol` vowel twice; it was already defined, and defining it twice will break the code and the phonemizer will not work (see oxygen's message; I think this is why the tests might be failing). Please fix that, thanks in advance.

stakira:
Sorry, it has been a while and there are conflicts. Could you resolve it?

PR #1208, opened 2024-07-06
[Bug fix; important] [JA Monophone G2P] Small phoneme fix

lottev1991:
I noticed I forgot to include two phonemes in the G2P; one of them is very rare, and not even used by many banks, but it's still an oversight.


lottev1991:
Apologies for the failing checks; I have no idea what's wrong.

PR #1217, opened 2024-07-20
Add .ustx file association on Windows and MacOS

RedBlackAka:
Added .ustx file association to the Windows installer and MacOS, so opening a file will automatically open it in OpenUtau.


OpenUtau/OpenUtau.csproj
stakira, line 1: Please try to not change the line ending...

RedBlackAka: I think I fixed my oversight now.

RedBlackAka: > Please try to not change the line ending...

I think I fixed my oversight now.

PR #1223, opened 2024-07-26
Add a 'reset all' editing macro

oxygen-dioxide:
The editing macro is called `Reset notes to default`. Please tell me if you come up with a better name.


overdramatic:
Maybe "Reset to Default" could be better?

PR #1240, opened 2024-08-13
ZH CVVC: Add tests for mixed hanzi and pinyin lyrics

oxygen-dioxide:
prevent bugs like what #1235 fixes from appearing again


oxygen-dioxide:
GitHub actions tests will pass after #1235 is merged

PR #1244, opened 2024-08-17
Add [DIFFS EN+] Phonemizer

Cadlaxa:
### `This phonemizer for Diffsinger uses ARPA+ G2p so the g2p features are the same with the ARPAsing+ Phonemizer`
- **G2p Features**
    - Support for `dr`, `tr`, `dx` ,`ax` and `q` (but `q` is replaced with `cl` and `vf` **(phonemizer feature)**)
    - `'` prefix on notes triggers glottal stop, if added to a vowel and the prevNeighbour is null, `vf` is used instead of `cl`
    - `'` suffix for Manual relaxed consonants, (applicable only if the word ends with `t` or `d`)
- **Phonemizer features**
    - Conditional replacements for `q` phoneme because the g2p uses `q` instead of `cl`
    - Automatic replacements of `t/d` to `dx`
    - Automatic replacements for `n dx` sequence and replace it with `n`, the actual phoneme for this is `nx` like `(winner [w ih nx er])`


Cadlaxa:
the recent pr is dependent on #1248 so it errored out

stakira:
tests failing?

PR #1245, opened 2024-08-18
Add Millefeuille French and Marzipan German Phonemizers

imsupposedto:
French is UFR Millefeuille and Marzipan German is a new system. For Marzipan it has a new phoneme system, hand written dict, and improved G2P performance over existing German phonemizer. All Millefeuille vbs will update with this phoneme system in the future as native integration and XLS for Marzipan within Millefeuille is currently being tested. Phonetic system created and owned by imsupposedto.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 OpenUtau is a free, open-source editor made for the UTAU community.
 
-[![AppVeyor](https://img.shields.io/appveyor/build/stakira/OpenUtau?style=for-the-badge&label=appveyor&logo=appveyor)](https://ci.appveyor.com/project/stakira/openutau)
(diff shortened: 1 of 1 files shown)


stakira:
Please try to avoid flipping line endings in files. You can see Github shows diff for every line of `LyricsHelper.cs` and `PhoneticAssistantViewModel.cs` just because the line ending change. If you use VS or VS Code, at the right bottom corner where shows CRLF or LF you should be able to change the line ending for file.

imsupposedto:
> Please try to avoid flipping line endings in files. You can see Github shows diff for every line of `LyricsHelper.cs` and `PhoneticAssistantViewModel.cs` just because the line ending change. If you use VS or VS Code, at the right bottom corner where shows CRLF or LF you should be able to change the line ending for file.

What line ending type does it need to be?

lottev1991:
@imsupposedto It should be CRLF.

imsupposedto:
> @imsupposedto It should be CRLF.

Hm, I hadn't changed it from that. Must be some weird Visual Studio thing. I'm correcting it now and will update the pull request

PR #1248, opened 2024-08-22
Support Diffsinger multi-dictionary

oxygen-dioxide:
This PR add support for diffsinger multi-dictionary (https://github.com/openvpi/DiffSinger/discussions/203)
- Support `phonemes.json` file, which is successor of `phonemes.txt`. (If the file name suffix is `.json`, it will be loaded as json format)
- Support language embedding

Currently, in phonetic hint (phoneme list surronded by `[]` in the lyric of the note), you don't need to include language prefix, and the phonemizer will assume that it uses the language of the current phonemizer. (Of course, you can also include the language prefix. You can use phoneme from another language in this way)

However, if you edit the phonemes manually in the phoneme panel at the bottom, you have to input the whole phoneme name with the language prefix.

image

PR #1255, opened 2024-08-26
Avoid repeated dictionary and model loading in DiffSingerBasePhonemizer

yqzhishen:
The former code loads the dictionary and duration model every time the phonemizer runs (include every single edit on the segments). This PR fixes this bug and the phonemizer no longer stucks.


yqzhishen:
There seems like a conflict between this PR and #1248 in the return type of method `DiffSingerBasePhonemizer._executeSetSinger()`. If one is merged, the other should be modified, otherwise the project will not compile.

PR #1262, opened 2024-09-01
 Remove restart notes from preferences where it doesn't apply and installer improvements

RedBlackAka:
Remove restart notes from preferences that do not actually need a program restart to change. Also made some adjustments to the installer.


RedBlackAka:
Note that the checks currently fail because of the security vulnerability.

PR #1268, opened 2024-09-03
Add DrawLine Pitch Tool

rokujyushi:
I added this because it was difficult to draw straight lines with the draw pitch tool.


rokujyushi:
image
I need input on where to add the tool.

rokujyushi:
image
The order of the icons has been changed.

PR #1269, opened 2024-09-04
Add shortcut for some Menu Item

JackZ2024:
add shortcut for some Menu Item


maiko3tattun:
Isn't Alt+V a duplicate?

JackZ2024:
Yes, it is indeed repeated, I have changed it to Alt+C

PR #1274, opened 2024-09-07
Added "ThaiG2p" & "DIFFS TH" DiffSinger Thai Phonemizer

spicytigermeat:
- Trained ThaiG2p on Thai Symbols, WER 0.0396 (99.96% of ~31K words were correctly guessed upon testing the G2p model, this G2p is very accurate! Any errors were only off by 1 or 2 phonemes maximum.)
- This G2p does use slightly different hyper-parameters and a very different optimizer for training, but in my experiments the results obtained are dramatically more accurate than the default parameters. The G2p.onnx model file is roughly ~14mb. I have not noticed this to cause a slowdown in OpenUTAU at all.
- Added the G2p to the Lyric Helper & Phonetic Assistant.
- Included a DiffSinger G2p Phonemizer for Thai. It reads "dsdict-th.yaml" and directly supports Thai characters!

Phonemes used for this G2p and Phonemizer are PrintMov's Phoneme system, which you can see [here](https://thaids.printmov.com/). There is a public dataset available for training on this dataset, so there is precedent for it to be included in OpenUTAU.

See example using Printto Magicbeat's Thai DiffSinger DB
image


stakira:
Please revert the 30+ files that do not have real changes.

mrtigermeat:
Can you please close this pull request? I lost access to my old account and cannot make changes to this one and have resubmitted the pull request. Thank you.

PR #1275, opened 2024-09-09
[DIFFS EN+] fixes to phoneme replacements with LangCode

Cadlaxa:


oxygen-dioxide:
Not all voicebanks use multi-dict. Old voicebanks and English-only voicebanks don't use multi-dict. 
I'll fix this for all g2p voicebanks in their base class.

stakira:
> Not all voicebanks use multi-dict. Old voicebanks and English-only voicebanks don't use multi-dict. I'll fix this for all g2p voicebanks in their base class.

What's the current status of this issue now? Is this PR still needed? Or is it better now?

Cadlaxa:
This pr fixes the issue with the language code update, cuz before it expects to use the multi-dict but now, it works without them (because of oxygen's update to phonetic hints without typing the language code)

PR #1281, opened 2024-09-10
Diffsinger Ja phonemizer: support kana

wolfgitpr:


oxygen-dioxide:
This PR adds a kana to romaji converter to DIFFS JA phonemizer, similar to DIFFS ZH's hanzi to pinyin converter. After this PR, voicebank developers only have to include romaji in their dsdict-ja.yaml, and the user will be able to input either romaji or kana as lyrics.

wolfgitpr:
Operational error leading to branch closure.

PR #1286, opened 2024-09-20
Fix LyricBox UI

maiko3tattun:
Fixed duplication of alias and folder paths when folders are deep.
image
↓
image


stakira:
Even though the lyric box before this change is suboptimal, this change hides the more useful part of the path, i.e., the name of the leaf subfolder, which is usually the voice color name. I'd prefer it hide the upper levels.

maiko3tattun:
Since it was not possible to hide the front of the path, I made it harder to cover it by erasing the singer's name and added a tooltip.
image

PR #1288, opened 2024-09-22
Remove support for outdated DiffSinger script files

komisteng:
When stakira/OpenUtau first became compatible with DiffSinger, the early versions of DiffSinger, along with its corresponding datasets and project file specifications, had already been replaced by newer versions.
OpenVPI completely removed compatibility with the old version's dataset in their fork of DiffSinger v2.3.0, and now almost no one should be using this outdated specification.
In OpenUtau, exporting old version DiffSinger scripts is labeled as 'Export DiffSinger Scripts To,' which can easily create the illusion that both the old and new versions are compatible with each other. I think this should be removed, or at the very least, marked somewhere as 'outdated.'
5951fb815c763b8acb8f7064f7a997df


stakira:
This PR was stuck at the tests and couldn't get merged earlier. Now it's having conflict. If you still want to merge it consider rebase it or resubmit.

PR #1300, opened 2024-10-13
Fixed BaseKoreanPhonemizer to use voice color

EX3exp:
- Fixed the bug that BaseKoreanPhonemizer was not appling voice color suffix.


EX3exp:
This bug fix has been applied in #1305, I'll close this PR.

PR #1305, opened 2024-10-18
Add Romanized lyrics support to all Korean phonemizers

2xxbin:
I added Romanized lyrics mapping to the Korean phonemizer.


2xxbin:
Added standard Korean Romanization and non-standard Romanizations used in the UTAU community.

PR #1309, opened 2024-10-19
Add Marzipan and Millefeuille phonemizers

imsupposedto:
Having another go at this, hopefully this time line endings are correct. I've ensured that visual studio was set to CRLF for the duration of my edits and used VS to download the repository instead of downloading like a zip from GitHub. Not sure what else it'd be if this doesn't fix it.


imsupposedto:
Checks fail due to global issue with system.text.json vulnerability

Cadlaxa:
Hi, if you pull for the latest branch on main, the net6 vulnerabilities will be fixed, ou moved to net8 a while ago because of the vulnerabilities

PR #1333, opened 2024-11-19
Move userdata from Documents to AppData for Windows installed version

RedBlackAka:
This changes to location of userdata ("OpenUtau" folder) from Documents to Appdata\Roaming for the Windows installed version, following good developer practices and cleaning up the experience for users. AppData is designed for that and most applications (including the original UTAU) follow this convention. This also includes a function that moves the old folder and all its contents into the new location, so users can update hassle free. Portable version is unaffected by this change. External singer paths are also unaffected.

Tested and works flawlessly. Only moves the OpenUtau folder if it detects it being used for userdata (by checking for prefs.json). Will likely remove this moving function in the future.

Additional note: this should also help users experiencing issues with OneDrive syncing their folders, as AppData is not a affected folder.


oxygen-dioxide:
Maiko has provided a PR that allow the user to choose where to store OpenUtau's data: https://github.com/stakira/OpenUtau/pull/1513

RedBlackAka:
There however still is a default path, so this still applies

PR #1342, opened 2024-11-28
Refactor preferences page to using a navigation sidebar, rearrange preference items

oxygen-dioxide:
There are more and more preference items. Before this change, we have to scroll down to find the preference item we want. After this changes, the preferences page uses a sidebar for navigation, like how windows settings app does.

Preferences items are also rearranged. The main change is that two new categories, "UTAU" and "DiffSinger" are created for renderer-specified preference items, because some users only use UTAU voicebank or only use DiffSinger voicebanks. Having "UTAU" and "DiffSinger" will make these preference items easier to find.
image


oxygen-dioxide:
Here are the preference items in each page:
image
image
image
image
image
image
image

oxygen-dioxide:
Reopen to rerun github action tests

PR #1345, opened 2024-11-30
When "saving as", or exporting wav, use the filename of the current file by default

oxygen-dioxide:
image


oxygen-dioxide:
Reopen to rerun github action tests

PR #1367, opened 2024-12-28
Automatic Wine Location For Resampler Script

ochakochatblanc:
Automatically finds the location of wine with 'which wine' command, otherwise defaults to '/usr/local/bin/wine32on64'.


ochakochatblanc:
im a goofy goober so idk if this code is good or not >.>

oxygen-dioxide:
https://github.com/stakira/OpenUtau/pull/1571 we are going to use wine directly from openutau instead of using a sh wrapper

stakira:
Superseded by #1571

PR #1368, opened 2024-12-28
Update simplified Chinese localization

oxygen-dioxide:


oxygen-dioxide:
Included in https://github.com/stakira/OpenUtau/pull/1373

PR #1384, opened 2025-01-19
Ignore unset lines in oto.ini

maiko3tattun:


OpenUtau.Core/Classic/VoicebankLoader.cs
```diff
@@ -469,6 +469,12 @@ static Oto ParseOto(string line, FileTrace trace) {
                 oto.Error = $"{trace}\nFailed to parse overlap. Format is {format}.";
                 return oto;
             }
+            if (!IsTest) {
+                if (oto.Offset == 0 && oto.Consonant == 0 && oto.Cutoff == 0 && oto.Preutter == 0 && oto.Overlap == 0) {
```
stakira, line 473: Isn't this still valid oto for the original UTAU?

maiko3tattun: If all parameters are zero, synthesis is not possible, which causes OU crashes in some engines. Therefore, for safety reasons, I suggest that we don't use these oto.

PR #1385, opened 2025-01-19
Simplify title on titlebar and in Apps and Features

RedBlackAka:
Simplify title on the titlebar on the main window and in Apps and Features for the installer, because version is not necessary to be shown there and it cleans up the presentation.

### DIFF
--- a/OpenUtau.nsi
+++ b/OpenUtau.nsi
@@ -44,7 +44,7 @@
 
 ; MUI end ------
 
-Name "${PRODUCT_NAME} ${PRODUCT_VERSION}"
+Name "${PRODUCT_NAME}"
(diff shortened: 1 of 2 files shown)


oxygen-dioxide:
I still think it's necessary to have version number on the title bar. We often see people who encounter issues when using diffsinger just because they aren't using the latest version. Having the version number on the title bar makes troubleshooting easier when the user posts a scrrenshot in the community.

stakira:
> I still think it's necessary to have version number on the title bar. We often see people who encounter issues when using diffsinger just because they aren't using the latest version. Having the version number on the title bar makes troubleshooting easier when the user posts a scrrenshot in the community.

As explained

PR #1386, opened 2025-01-20
show singer default phonemizer in track phonemizer menu

oxygen-dioxide:
image


oxygen-dioxide:
@RedBlackAka Do you mean that you don't like this feature? 
My idea is: voicebank developer can declare a recommended phonemizer. If the user switched to another phonemizer accidentally or for certain reason, and want to go back to the original phonemizer, before this PR, they have to find the phonemizer manually in the list, or open character.txt by hand to see what's the original phonemizer. After this PR, they will be able to switch back to the phonemizer declared in the voicebank quickly.
If you don't like it, we can add a preference item.

RedBlackAka:
It is fine, initially though it would guess a phonemizer, but mixed it up with one of the suggested ideas.  I wanted to only have it shown if one was defined, which seems to be the case here.

PR #1396, opened 2025-01-25
Add RandomizeTiming function to batch edit

maiko3tattun:
Randomize note timing
image

Randomize phoneme offset
image


stakira:
Why not keep notes unchanged and randomize phonemes in same note using the same offset? You would be able to revert to standard timing reliably any time. I.e., randomize phoneme offsets by notes vs. randomize phoneme offsets individually.

maiko3tattun:
This PR contains two types of moves: one to move the note and one to move the offset.
The method of moving the offset is easier to manage, but does not change the pitch, so the nuance of doubling is not present. To reproduce more realistic doubling, the entire note must be moved.
This PR allows us to use both of these methods.

stakira:
oops conflict

PR #1400, opened 2025-02-01
Improvement of Knife Tool and bug fixes for Tool+Ctrl key

maiko3tattun:
### New Feature
- Default lyrics for notes splitted by the Knife Tool can now be specified.
image

### Bug Fix
- All tools except the Cursor Tool can now select notes while Control/Command is held down.
- Fixed a bug where notes could not be selected when Control/Command and another modifier key were simultaneously held down.
- Allows deselection of notes when right-clicking with the Knife Tool.


maiko3tattun:
The length of the note cut with the knife tool can now be changed.

https://github.com/user-attachments/assets/74618416-0925-4ae6-b643-71bcd20ea6c6

stakira:
"Fixed a bug where notes could not be selected when Control/Command and another modifier key were simultaneously held down."
I don't think it should work that way. It's easy to create conflicts between operations. Each modifier key combo should be exact.

maiko3tattun:
Reverted to the specification that notes can be selected when the modifier key is only Control/Command.

OpenUtau/Views/PianoRollWindow.axaml.cs
```diff
@@ -593,18 +587,17 @@ private void NotesCanvasLeftPointerPressed(Control control, PointerPoint point,
...
+                    editState = new NoteSplitEditState(
+                            control, ViewModel, this, noteHitInfo.note);
                 } else {
                     editState = new NoteMoveEditState(control, ViewModel, this, noteHitInfo.note);
                     Cursor = ViewConstants.cursorSizeAll;
                 }
                 return;
             }
-            if (ViewModel.NotesViewModel.CursorTool ||
-                ViewModel.NotesViewModel.PenTool && args.KeyModifiers == cmdKey ||
```
stakira, line 603: Why change these?

stakira: Oh I see.

stakira (approved):

PR #1418, opened 2025-02-13
Add option to enable/disable logging

RedBlackAka:
Adds an option in Preferences to enable/disable logging to file, for users that do not like files building up. Enabled by default.

Tested on Windows and Mac, and works flawlessly. Only creates files when enabled. Changing the setting requires a restart.

Screenshot 2025-02-13 at 22 19 48


stakira:
I don't think we need this.

oxygen-dioxide:
I don't think we should allow users to disable logging completely. If PR author doesn't like having a lot of log files in their Logs folder, "only keeping logs of recent 30 days" is better.

PR #1428, opened 2025-02-28
Microtonal tuning function and batch editing of randomization

maiko3tattun:
## New features
- Edit the pitch in 1cent increments.
image
- Randomize a batch of microtones. This is useful for creating choruses and doublings.
image


oxygen-dioxide:
I think it would be better if tuning is applied before pitch control points, which means that pitch control points will snap to note tone adjusted with microtone. 
I implemented this in https://github.com/oxygen-dioxide/OpenUtau/tree/microtune . My branch also has the following features implemented:
- The position of the note shows its tuning
- diffsinger pitch model takes tone with tuning as input

yqzhishen (dismissed):

oxygen-dioxide (dismissed):

PR #1435, opened 2025-03-04
[Updated Pull Request] Add ThaiG2p & DIFFS TH Phonemizer.

mrtigermeat:
- Add ThaiG2p, trained on custom hparams for a very good Thai G2p (99% accurate!!)
- Add DIFFS TH Phonemizer.


stakira:
Adding g2p-th.zip twice?

PR #1447, opened 2025-03-14
Removed 0.9f limit of Preutter delta

Cadlaxa:
This removes the 0.9f preutter delta cap so the preutter length/duration can go all the way to the previous preutter just like the original utau mechanics (also it will not exceed beyond the previous preutter)


Cadlaxa:
image

OpenUtau.Core/Ustx/UPhoneme.cs
```diff
@@ -118,9 +118,9 @@ void ValidateOverlap(UProject project, UTrack track, UPart part, UNote note) {
                     autoPreutter = maxPreutter;
                     autoOverlap *= ratio;
                 }
-                if (autoPreutter > prevDur * 0.9f && overlapped) {
-                    double delta = autoPreutter - prevDur * 0.9f;
-                    autoPreutter -= delta;
+                if (autoPreutter > prevDur) {
```
stakira, line 121: Why remove the overlapped check?

AnAndroNerd: The overlap check breaks a whole bunch of voice banks, ie VCCV and anything OTO'd like it. Here's a small video that explains it [](https://youtu.be/F92r7YYy3-U?si=CWCQOS7yhAbq-Vtc)

stakira: I still don't get it. The overlapped check is about whether two notes are consecutive or not. It shouldn't affect consecutive notes.

maiko3tattun:
It is the opinion of the Japanese UTAU community that it is not a good idea to have the preutter automatically on the far left, since it generally takes several tens of milliseconds from the beginning of the note for the pronunciation to stabilize.

This should be handled in the “lengthen crossfade” batch edit, and adjusting this length is part of the PR I am working on.

wik-wav:
The overlap going all the way back to the previous preutter is also very important for long consonants in arpasing, eg. if any instance of [ch] appears, it will be covered by the previous note's loop that's too far into the VC, which results in this consonant either not making any sound at all on really tight clusters, or having its onset covered (making it more difficult to understand the phrase, because it'll sound like a different consonant entirely)
Removing the limit will also make consonant clusters sound much better without manual edits because really short VCs won't get stretched unnecessarily. This stretching I'm trying to describe (that is a result of the overlap limit Cadlaxa tried to unlock) is very obvious when using certain resamplers like tips with arpasing voicebanks.

PR #1450, opened 2025-03-15
Change the UI name in "Edit Subbanks" in the Singers menu

AnAndroNerd:


stakira:
Strictly speaking, the tone range subbanks are not voice colors...

AnAndroNerd:
This is true, but I do think this way of saying it is more intuitive.

PR #1454, opened 2025-03-16
Add Singer Settings Dialog

maiko3tattun:
## New Feature
Function to edit character.yaml
It is easy to set localized names, etc., which have been difficult to set in the past.
image

Note:
In conventional UTAU, character.txt, unlike oto.ini, does not allow specifying text encoding and may not be readable if the environment (OS, language, etc.) changes.
Therefore, the text encoding of character.txt can be selected separately.
Also, the same content was also written in character.yaml, assuming that character.txt may not be readable.

## Changes
- For compatibility with UTAU, “other information” can also be described in character.yaml.
- The field in VoicebankConfig is now nullable, since it used to be actually assigned null.


Cadlaxa (approved):
Tested this and it works fine, needs some fixes for the strings.axml since it got conflicts in the main

nagotown:
not working when merged with main on my end

PR #1457, opened 2025-03-21
[Waiting for test] Show window title in macOS

maiko3tattun:
Changes: ExtendClientAreaToDecorationsHint  is now always false except for splash screen.

Fixed that version, file paths, etc. are not displayed on macOS.
However, I have not been able to test it on an actual device.
Can anyone give me feedback? Or if there is a reason why it shouldn't be modified, please let me know.


RedBlackAka:
Tested and seems to work for every window 👍

Issue #1459, opened 2025-03-22
OpenUtau's data folder appeared in the wrong place on macOS


Nyaacinth:
### Acknowledgement

- [x] I have read Getting-Started and FAQ

### 🐛 Describe the bug

Like what the title said, OpenUtau's data appeared in `/Users/<username>/Documents/Library/OpenUtau` and `/Users/<username>/Documents/Library/Caches/OpenUtau`.
That's because of a mistake in `OpenUtau.Core/Util/PathManager.cs` which used `Environment.GetFolderPath(Environment.SpecialFolder.Personal)` to get user's home directory.
But that actually ended up with Documents folder, simply replace it with `Environment.GetFolderPath(Environment.SpecialFolder.UserProfile)` will get the previous behavior.

### Explains how to reproduce the bug

1. Download or build the latest beta
2. Launch it
3. You will get those folders

### OS & Version

macOS 15.3.2, OpenUtau 0.1.549

### Logs

(Not related, and no log associated in logfile)

Nyaacinth:
If necessary, I can do a pr to fix this. But maybe you'll want to test it more comprehensively.

PR #1460, opened 2025-03-22
Fix OpenUtau's data folders appearing in the wrong place on macOS

Nyaacinth:
Ref: https://github.com/stakira/OpenUtau/issues/1459

By using `Environment.SpecialFolder.UserProfile` instead of `Environment.SpecialFolder.Personal` to get user's home directory, this will fix the issue that OpenUtau's folder appearing in `/Users/<username>/Documents/Library` but not `/Users/<username>/Library`.

Since I don't have a matrix to fully test this on a variety of macOS devices, you might want to test it more comprehensively before merging or pushing it to release.

### DIFF
--- a/OpenUtau.Core/Util/PathManager.cs
+++ b/OpenUtau.Core/Util/PathManager.cs
@@ -16,7 +16,7 @@ namespace OpenUtau.Core {
         public PathManager() {
             RootPath = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
             if (OS.IsMacOS()) {
-                string userHome = Environment.GetFolderPath(Environment.SpecialFolder.Personal);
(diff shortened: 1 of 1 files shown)


RedBlackAka:
Could you potentially add a temporary function in the main app startup to move the OpenUtau folder and its contents if detected, so that existing users can update hassle-free? 

On a note: this is similar to my Windows AppData PR and a good cleanup for following conventions

Seele-Vollerei32:
Due to [this change on dotnet](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/getfolderpath-unix), all the UNIX like os are affected, so please change the same field of linux

Seele-Vollerei32:
However, there is a same problem under Linux and haven't been fix. Maybe I need to make another PR

PR #1468, opened 2025-03-27
Allow assignment of an oto set to multiple colors

maiko3tattun:
Until now, when multiple colors referenced the same oto set, the fallback made it look like it was working correctly, when in fact it was not. This voice bank is now officially supported.
(This did not work correctly for phonemizers that support multiple oto types like JA VCV&CVVC, because sometimes it was necessary to dare not fall back on them.)

prev (Only the CV falls back, not the VC)
image

now (Both CV and VC are recognized as the correct color)
image


maiko3tattun:
Discord ideas thread: https://discord.com/channels/551606189386104834/1323319504393998377

maiko3tattun:
Fix color column
image

PR #1472, opened 2025-03-29
Improvements preferences and related to singers folder

maiko3tattun:
## Changes:
- Added “General” to the Preferences dialog and organized the items.
image
- The Preferences dialog now allows us to see and open the Singers folder.
image
- Improved behavior of track headers when no singer is installed.
image

## Internal Changes:
- Rearranged PreferencesViewModel items for easier reading.


RedBlackAka:
Please reconsider if/how you do the "General" option, it splits out settings that currently are in appropriate sections and clutters the list with an underutilized option. If going that way, keep all the settings in mind and for example put all Appearance settings as a section under General etc.

RedBlackAka:
I suggest simply reorganizing Appearance to be on top instead, making two of these options immediately appear for users without the need for another tab.

maiko3tattun:
First of all, considering that people cannot read "Appearance", it is imperative that the language settings come first.
I think language and UI customization are too different in priority, but simply moving the Appearance tab to the top is not a bad idea.

Beta is sometimes recommended to avoid bugs when providing user support, so I want to place it where beginners can easily find it. I would like to hear everyone's opinion on this.

PR #1477, opened 2025-04-02
Add Style manager＿

rokujyushi:
Because of problems with fonts and styles on MacOS, the display of controls is corrupted.
The styles to be loaded for each operating system have been separated.


stakira:
closing since we have custom theme and theme editor now

PR #1480, opened 2025-04-03
Fixed issue with voice bank loading due to differences in spec between OS

maiko3tattun:
Fixed wav not loading when one of the wav file name and oto specified file is NFD and the other is NFC.
- This was occurring when a voicebank created on macOS or smartphone was brought to Windows, or when a zip created on Windows was extracted on a mac.
- I have tested each of the internal and external wavtools and engines.


maiko3tattun:
This only works when a wav is not found, so it does not affect the voicebanks that have been working so far :kissing_heart:

PR #1483, opened 2025-04-05
Clear caches for each phrase

maiko3tattun:
## New Feature:
- Keep the cache filenames in each phrase and delete the files of phrases that contain the selected notes (partial filename match).
image

## Changes:
- Translate progress bar text

## Note:
I have tested with Classic Singer, but I am not fully aware of the specifications for the other renderers and may not be able to erase all caches.


yqzhishen (commented):
DiffSinger caches many intermediate variables if the tensor cache is enabled. So if these cache files are not deleted, it still generates the same result on next run. Maybe you should call `Delete()` on the DiffSingerCache instances. (ref: OpenUtau.Core/DiffSinger/DiffSingerCache.cs)

maiko3tattun:
I tried to be able to delete the DiffSinger cache. However, I could not tie the durationCache to the phrase (´･ω･｀)

PR #1484, opened 2025-04-05
Fixed progress bar in pianoroll and minor visual issues

maiko3tattun:
## Fixes:
- Fixed progress bar not showing in pianoroll window (PianoRollViewModel)
- Fixed that track numbers were shifted to the left by certain operations (TrackHeader)
- Adjust margin in note properties panel (NotePropertiesControl)

## Internal Changes:
- Removed unnecessary nesting of panels (MainWindow, PianoRollWindow)


maiko3tattun:
@stakira I would appreciate it if you could merge this as soon as possible.
The MainWindow and Pianoroll are hard to maintain because git misleads it into thinking it is changing a lot of lines when in fact it is only rewriting a few.

stakira:
I'd say no need to remove the nesting just for the sake of it. Helps keep git history clean.

PR #1487, opened 2025-04-10
Seperated steps for DiffSinger

lunaiproject:
Added ability to set different step values for DiffSinger models: **Acoustic**, **Variance** and **Pitch**
This will speed up the rendering and give more options.

According to my experience, only **5** steps are enough to render a good Pitch, while Variance parameters require a value around Acoustic, but can still be much lower if it doesn't contain complex parameters. So I suggest leaving it initially the same as Acoustic, so that the user can change it on their own if they see fit.

image


oxygen-dioxide (dismissed):

yqzhishen (changes requested):
I would argue a default **10 steps** for pitch model inference. Maybe 5 is a little too radical.

keirokeer:
Ok!

yqzhishen (approved):

stakira (approved):

PR #1488, opened 2025-04-10
Ctrl+R for Load rendered pitch

lunaiproject:
This will make it faster and more convenient to generate new pitch takes by selecting notes on Piano Roll


keirokeer:
Now it's referred to Notes > Load Rendered Pitch and should solve the problem

oxygen-dioxide (approved):

PR #1492, opened 2025-04-13
When saving oto.ini, always include alias in oto lines even if the filename and alias are same

oxygen-dioxide:
If the wav file and oto.ini file are in a subfolder, for an oto line without alias, on Ameya UTAU, user have to input`<foldername>/<wavname>` as lyrics, which is incompatible with existing USTs, and is hard to use


yqzhishen (commented):
I'm not familiar with these modules 😓

maiko3tattun:
Nowadays it is more common to fill in the alias even if it is the same as the wav file name👍

PR #1503, opened 2025-04-20
Fixed Vibrato VolLink was not copied

maiko3tattun:
## Fixes:
- Fixed vibrato volume link not being copied when PasteSelectedParams.
- Fixed volume link not being reset by "Reset vibratos" in batch edit.

## Improvments:
- By merging the commands, PasteSelectedParams processing has been sped up.

### DIFF
--- a/OpenUtau.Core/Commands/NoteCommands.cs
+++ b/OpenUtau.Core/Commands/NoteCommands.cs
@@ -182,6 +182,30 @@ namespace OpenUtau.Core {
         };
     }
 
+    public class SetVibratoCommand : VibratoCommand {
+        readonly UNote note;
+        readonly UVibrato newVibrato;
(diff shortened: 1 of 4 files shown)


OpenUtau.Core/Commands/NoteCommands.cs
```diff
@@ -182,6 +182,30 @@ public VibratoCommand(UVoicePart part, UNote note) : base(part, note) { }
         };
     }
 
+    public class SetVibratoCommand : VibratoCommand {
```
stakira, line 185: Might as well just use SetVibratoCommand for all vibrato changes.

maiko3tattun: Not all vibrato changes could use SetVibratoCommand, but some commands could be replaced with it to speed up the process.

PR #1507, opened 2025-04-22
Add pkg

rokujyushi:
Supports installation by package and avoids the procedure of running commands from the terminal.

It is no longer necessary to enter the following commands.
```
xattr -rc /Applications/OpenUtau.app
```

I think there is a reason why it was provided in a dmg file, so I have kept the instructions for creating the dmg file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -138,14 +138,14 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
-          name: OpenUtau-osx-${{ matrix.arch.name }}.dmg
-          path: OpenUtau-osx-${{ matrix.arch.name }}.dmg
+          name: OpenUtau-${{ matrix.arch.name }}.dmg
+          path: OpenUtau-${{ matrix.arch.name }}.dmg
         if: ${{ !inputs.release && matrix.arch.os == 'osx' }}
 
       - uses: actions/upload-artifact@v4
         with:
-          name: OpenUtau-osx-${{ matrix.arch.name }}.pkg
-          path: OpenUtau-osx-${{ matrix.arch.name }}.pkg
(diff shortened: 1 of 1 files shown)


rokujyushi:
It provides a new option for the installation method and does not intend to erase the dmg file installation method.  
I think the pkg file is a better choice for some users than the dmg file because both have advantages and disadvantages.
I think the pkg file will be for users who use a Mac but are not familiar with Terminal or commands.

## dmg file.
### Advantages.
- Traditional method
- Fully compatible with automatic updates.
### Disadvantages.
- Requires command input from Terminal (difficult for users who use a Mac but are not familiar with Terminal or commands)
- Security permissions need to be granted for dmg and app files.

## pkg file.
### Advantages.
- Security permissions only once
- No need to enter commands from Terminal
### Disadvantages
- Administrator permission required when updating.
- Partial support for automatic updates (it even downloads the pkg file automatically, but you need to follow a few steps to update it).

rokujyushi:
@stakira 
It provides a new option for the installation method and does not intend to erase the dmg file installation method.
I think the pkg file is a better choice for some users than the dmg file because both have advantages and disadvantages.
I think the pkg file will be for users who use a Mac but are not familiar with Terminal or commands.

## dmg file.
### Advantages.
- Traditional method
- Fully compatible with automatic updates.
### Disadvantages.
- Requires command input from Terminal (difficult for users who use a Mac but are not familiar with Terminal or commands)
- Security permissions need to be granted for dmg and app files.

## pkg file.
### Advantages.
- Security permissions only once
- No need to enter commands from Terminal
### Disadvantages
- Administrator permission required when updating.
- Partial support for automatic updates (it even downloads the pkg file automatically, but you need to follow a few steps to update it).

PR #1509, opened 2025-04-26
Add welcome page

oxygen-dioxide:
This PR adds a welcome page to OpenUtau, like visual studio and office.
Welcome page provides an elegent way to show warnings and infos on startup, without popping dialogs one by one. It will also benefit mobile user experience, because on mobile we usually open the app first, and open files from app, unlike on desktop where we open files from folder.
Users can turn it off in preferences, if they prefer the previous behaviour where OpenUtau creates a new project on startup

Implemented features:
- Create new project
- Open project
- Recent projects
Todo:
- Recovery backup
- Startup warning (such as non-ascii path)
image


XieYimin3 (approved):
approval

stakira:
It seems unnecessary to add a preference though. The new project is just one click away.
Also... how does the on-by-default tips overlay work on the first launch?

PR #1513, opened 2025-04-29
Custom Data Path

maiko3tattun:
## Development Background: 
- In Windows, placing data in the Documents folder causes various troubles.
- There is a demand to specify the location of folders with large data size such as Singers, Cache, and Resampler.
- Confusion arises during user support because the data path changes depending on whether the installer was used or not.

## Improvements:
- Data path can now be changed/checked in preferences.
- Check if the custom data path can be written to at startup and when the path is changed. If not authorized, fall back to the original data path.
- The location of prefs.json has not been changed.
- For now, it will not automatically move data when a path is changed.


maiko3tattun:
I understand that there are multiple opinions regarding the data path. Please consider this one of the suggested solutions.

oxygen-dioxide:
I think we can follow what zotero does.
After changing data path, zotero won't autometically move the folder. It will show a dialog to tell user about this and exit. 
image
image
image

OpenUtau should exit after changing data path, and use the new data path after the user launches OpenUtau the next time. because if data path is changed but the software is still running, voicebanks and resamplers may not work correctly.

oxygen-dioxide:
Also users can use the changeable data path to separate their singers into multiple workspaces, because the more singer user installed, the slower OpenUtau launches.

maiko3tattun:
- A message is displayed when the custom data path is changed.
- Logs, Note Presets, and Backups are not moved to the custom data path.
image

maiko3tattun:
- After selecting a custom data path, the dialog will no longer appear if the path remains unchanged
- Fixed a bug where errors during preference loading were not logged

rokujyushi:
最新環境とマージしてPRテストを完了させてください。
Please merge with the latest environment and complete the PR testing.

maiko3tattun:
image
I've added folder permission checking at startup. I've also added a cloud storage check, which has been a source of problems for many users.

PR #1516, opened 2025-04-30
Toggle playback line

keirokeer:
Adds the ability to snap the playback line to the top arrow with a stable width.
You may need this if you need to record a nice looking video with a smooth playback line.

_It can probably be done better, so I'd love some help! However, it works._


oxygen-dioxide (approved):

oxygen-dioxide:
I think preferences is a better place for this than toolbar

PR #1517, opened 2025-04-30
Edit Waveform Image

keirokeer:
If the peak is less than 0.01 pixels, it will be empty. This way the phrases will not be on the same line, but separated. This will change the visual of the waveform more conveniently.


stakira:
I don't understand what problem this PR is trying to solve.

keirokeer:
This will remove the line on the waveform that stretches across the entire piano roll. I'd like it to be removed permanently, but I can do it as an optional feature in the preferences if you see fit.

image
image

OpenUtau/Controls/WaveformImage.cs
```diff
@@ -143,9 +145,10 @@ private void DrawPeak(int[] data, int width, int x, int y1, int y2) {
                 y2 = y1;
                 y1 = temp;
             }
-            for (var y = y1; y <= y2; ++y) {
-                data[x + width * y] = color;
-            }
+            if (y2 - y1 > 0.01) {
```
stakira, line 148: Are you sure this compile?

PR #1521, opened 2025-05-04
Hide DiffSinger Language Code in Phoneme Canvas

mrtigermeat:
This pull request will add the functionality to optionally obfuscate the DiffSinger language prefix in the phoneme editor. I have personally found this to be an annoyance while using Multi-Dictionary DiffSinger models. DiffSinger recently adopted Multi-Dictionary as the standard for voice library creation, and as such this issue will be present in all DiffSinger models.

This pull request does the following:
- Changes the `GetLangCode()` function in all applicable DiffSinger phonemizers (G2p & Base) to be a public method instead of private, so it can be accessed in `PhonemeCanvas.cs` . This may cause incompatibilities with existing plugin phonemizers, but will not cause any issues with the built-in experience.
- Adds a new toggle switch under the DiffSinger section of the preference window to turn this feature on or off. It is turned off by default.
- Removes the language prefix from the text of the phoneme in the for loop in `PhonemeCanvas.cs`, but only for the language prefix of the selected phonemizer (hence why I had to change `GetLangCode()` to be a public method.)

Example:
ezgif-4c3070dbc39436

### DIFF
--- a/OpenUtau.Core/DiffSinger/DiffSingerBasePhonemizer.cs
+++ b/OpenUtau.Core/DiffSinger/DiffSingerBasePhonemizer.cs
@@ -30,7 +30,7 @@ namespace OpenUtau.Core.DiffSinger
 
         string defaultPause = "SP";
         protected virtual string GetDictionaryName()=>"dsdict.yaml";
-        protected virtual string GetLangCode()=>String.Empty;//The language code of the language the phonemizer is made for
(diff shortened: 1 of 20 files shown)


oxygen-dioxide:
There is a bug: when two phoneme aliases are too close, when hovering or double clicking on the second phoneme, it's the first phoneme that reacts
I fixed it in https://github.com/mrtigermeat/OpenUtau/pull/1

oxygen-dioxide (approved):

PR #1525, opened 2025-05-06
Add support for SVP (Synthv project file)

Cadlaxa:
native support for svp


OpenUtau.Core/Format/Formats.cs
```diff
@@ -16,6 +16,7 @@ public static class Formats {
         const string midiMatch = "MThd";
         const string ufdataMatch = "\"formatVersion\":";
         const string musicxmlMatch = "score-partwise";
+        const string svpMatch = "\"version\":";
```
stakira, line 19: This is too generic to be used to match a format. Try find something more specific.

Cadlaxa:
help is highly appreciated; there's a bug currently with the note onset and duration regarding their tempo also there's no parser for the curve params yet

Wisperer:
Been testing this. aside from a few quirks with SV2 2.2.1 it work really well.

PR #1527, opened 2025-05-06
Voicevox refactoring and additional functionality.

rokujyushi:
#### PR Classification
Changes aimed at adding new features and improving existing ones.

#### PR Summary
New functionality has been added to classes related to speech synthesis and existing functionality has been improved.  
- `RenderPhrase.cs`: a `phonemizerTag` field has been added and is now initialised in the constructor.  
- `F0Smoother.cs`: a new class has been added for smoothing F0.  
- `VoicevoxPhonemizer.cs`: the `SetSinger` method has been modified to determine the vowel extension note of a syllable.  
- `VoicevoxRenderer.cs`: The `Render` method has been significantly modified to manage speech synthesis parameters as `VoicevoxSynthParams`.  
- `VoicevoxUtils.cs`: new methods for querying F0 and volume have been added.


oxygen-dioxide:
By introducing `phonemizerTag` to `RenderPhrase`, you want to allow VoicevoxRenderer to know if it gets phoneme-level input or syllable-level input. I don't think this is a good idea because this will block potential third-party phonemizers, or users who manually place phonemes with `DEFAULT` phonemizer.

rokujyushi:
@oxygen-dioxide 
> By introducing `phonemizerTag` to `RenderPhrase`, you want to allow VoicevoxRenderer to know if it gets phoneme-level input or syllable-level input. I don't think this is a good idea because this will block potential third-party phonemizers, or users who manually place phonemes with `DEFAULT` phonemizer.

We have eliminated the `phonemizerTag` discrimination and separated the behaviour based on the content of the `RenderPhrase`.

Voicevox currently only accepts hiragana or katakana lyrics, so we are wondering if we should separate `VoicevoxRenderer` into a function that passes lyrics to Voicevox for synthesis and a function that passes lyrics and phonemes to Voicevox for synthesis.

Issue #1518, opened 2025-05-01
Cache resampler/wavtool options


shuntia:
### Acknowledgement

- [x] I have read Getting-Started and FAQ

### Description of the new feature / enhancement

It takes a very long time to load the resamplers menu after choosing the CLASSIC renderer.

### Proposed technical implementation details

cache the result of the recursive find in a static Mutex or Lazy

oxygen-dioxide:
It opens immediately on my side.
How many resamplers do you have? Also please provide your log file (you can find it with "Help -> Open Logs Location")

shuntia:
It seems to tree the whole resamplers directory and list out all the executable files.
I added hifisampler and it massively increased the time to load the resamplers, since it seems to have traversed down the python venv.
To be honest if I just move the "hifisampler" folder outside the resamplers folder, it should fix

[log20250513.txt](https://github.com/user-attachments/files/20199276/log20250513.txt)

PR #1535, opened 2025-05-14
Apply 'Load all depth folders' to resamplers, wavtools, dll phonemizers and UTAU plugins too.

oxygen-dioxide:
related issue: #1518

After this PR, the "Load all depth folders" setting item will also be applied to resamplers, wavtools, dll phonemizers and UTAU plugins. This will greatly increase OpenUtau's startup speed if the user has plugins or resamplers with many DLLs and complex path structure, such as hifisampler

I don't know if this should be a separate setting item from the current "Load all depth folders" for singers. feel free to share your idea.

### DIFF
--- a/OpenUtau.Core/Classic/PluginLoader.cs
+++ b/OpenUtau.Core/Classic/PluginLoader.cs
@@ -5,15 +5,23 @@ using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using Serilog;
+using OpenUtau.Core.Util;
 
 namespace OpenUtau.Classic {
     class PluginLoader {
         public static Plugin[] LoadAll(string basePath) {
             Directory.CreateDirectory(basePath);
             var encoding = Encoding.GetEncoding("shift_jis");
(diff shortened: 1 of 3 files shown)


stakira:
Checking dlls could be slow. `LibraryLoader.IsManagedAssembly()` was added to speed this up but probably not enough. Are you sure merely searching `plugin.txt` and `.exe` files matter?

maiko3tattun:
I don't know how this matter will be settled, but I personally would like to see it separated from the voice bank setting.

oxygen-dioxide:
> Checking dlls could be slow. `LibraryLoader.IsManagedAssembly()` was added to speed this up but probably not enough. Are you sure merely searching `plugin.txt` and `.exe` files matter? I have a feeling that all we need is to not read into Python runtime folder.

For UTAU plugins, we're already only searching plugin.txt. The problem is both UTAU plugins and external DLL phonemizers are installed into `Plugins` folder, so if there is a UTAU plugin with many DLLs, especially when it is also developed with dotnet, it will slow down the loading process of DLL phonemizers.

PR #1539, opened 2025-05-16
Support displaying real curves from DiffSinger variance predictor

yqzhishen:
## Motivation
DiffSinger acoustic models take actual expression values (rather than delta values) as inputs.
However, as OpenUTAU is originally designed for delta-like expressions, DiffSinger users has been suffering for a long time since they have to tune delta values on an invisible "real curve". Though it seems quite complicated to allow direct editing to the real curve, it could be a positive step to just display it on the expressions panel.

## Changes
This PR allows user to display and refresh the real curves on their expressions panel. Major changes are:
- `UCurve` class got two new fields, `realXs` and `realYs`. These fields are YAML-ignored.
- `MergedSetCurveCommand` got a new flag parameter `setReal` to indicate whether to update the delta values or the real values
- A new batch edit `RefreshRealCurves` to update `realXs` and `realYs` in `UCurve` was added.
- `IRenderer` interface got a new property `SupportsRealCurve` (defaults to `false`) and a new method `LoadRenderedRealCurves` (returns empty list by default). Only `DiffSingerRenderer` overrides them currently.
- New brushes were added, including a brush (and pen) to stroke the real curve and a linear gradient brush to fill the area under it.
- `ExpressionCanvas` got a new reactive property `ShowRealCurve`, in order to not render the real curve of the secondary key. Drawing logic was added here.

Some details:
- `LoadRenderedRealCurves` should return a list of `RenderRealCurveResult` objects. `ticks` is the relative tick to the phrase position (same as `RenderPitchResult`). `values` should be float values normalized between 0 and 1. Negative values are internally considered as breakpoints in the drawing logic.
- Because `MergedSetCurveCommand` only accepts integers, the drawing has a quantized precision of 0.001.

## Usage and demo
This feature is available only when DiffSinger tensor cache is enabled. The new batch edit "refresh real curves" generates or loads the real curves, apply the delta values edited by user on them, and displays the final model input on the background. The real curves will not refresh automatically, so user should run the batch edit manually to sync them.

image

### DIFF
--- a/OpenUtau.Core/Commands/ExpCommands.cs
+++ b/OpenUtau.Core/Commands/ExpCommands.cs
@@ -368,11 +368,11 @@ namespace OpenUtau.Core {
                 curve = new UCurve(descriptor);
                 Part.curves.Add(curve);
             }
-            getCurveXs(curve)!.Clear();
-            getCurveYs(curve)!.Clear();
+            getCurveXs(curve)?.Clear();
+            getCurveYs(curve)?.Clear();
             if (newXs != null && newYs != null) {
-                getCurveXs(curve)!.AddRange(newXs);
-                getCurveYs(curve)!.AddRange(newYs);
+                getCurveXs(curve)?.AddRange(newXs);
+                getCurveYs(curve)?.AddRange(newYs);
             }
         }
         public override void Unexecute() {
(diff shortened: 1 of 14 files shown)


XieYimin3 (approved):

XieYimin3 (commented):
Hoping the real curve could be editable soon!

PR #1544, opened 2025-05-19
[DiffSinger Phonemizers] Parse same phonetic result on lowercase and mixed lowercase/uppercase lyric notes

lottev1991:
This is an implementation similar to #1209 but only for DiffSinger phonemizers.

The default G2P behavior is that uppercase and lowercase lyrics give different results. This is intentional and has not been changed. However, DiffSinger users generally prefer to have the same phonetic result regardless of capitalization (akin to the UTAU phonemizers), especially when there's a capital letter at the beginning of a word (e.g. the beginning of a sentence, or languages such as German where all nouns are capitalized). As a result, I changed this behavior for DiffSinger phonemizers only.

I've created an exception for words that are all uppercase (such as abbreviations); this includes the global DiffSinger phonemes `SP` and `AP` (you'll still need to define them manually in your G2P-based `dsdict`s). This is because distinguishing all-caps words might be preferred in these cases. (Feedback on this is appreciated, however.)

### DIFF
--- a/OpenUtau.Core/DiffSinger/DiffSingerBasePhonemizer.cs
+++ b/OpenUtau.Core/DiffSinger/DiffSingerBasePhonemizer.cs
@@ -158,6 +158,21 @@ namespace OpenUtau.Core.DiffSinger
                 .ToArray();
         }
 
+        /// <summary>
+        /// Check if lyrics are all uppercase.
+        /// If true, do not parse as all lowercase.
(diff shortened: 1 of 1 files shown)


lottev1991:
I could see if that's possible; it might be. I believe I've tried it before and couldn't get it to work, but I'm willing to try again.

lottev1991:
Think I'll just close this.

PR #1550, opened 2025-05-24
Fixed unexpected behavior of welcome page

maiko3tattun:
## Fixes:
- Fixed no page transition when opening a project
- Disable ineffective menu on the welcome page

### DIFF
--- a/OpenUtau.Core/Util/Preferences.cs
+++ b/OpenUtau.Core/Util/Preferences.cs
@@ -188,7 +188,6 @@ namespace OpenUtau.Core.Util {
             public bool RememberMid = false;
             public bool RememberUst = true;
             public bool RememberVsqx = true;
-            public int LaunchBehaviour = 0;
             public int ImportTempo = 0;
             public string PhoneticAssistant = string.Empty;
             public string RecentOpenSingerDirectory = string.Empty;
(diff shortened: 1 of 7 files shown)


stakira:
What about:
1. We keep it simple and disable "File", "Edit" and "Project" menus entirely.
2. Remove recovery popup and move them to the welcome page. We will also be able to keep multiple recoveries this way.
3. Get rid of the preference of disabling welcome page. This way we can always show the recoveries.

Just a general thought: it's good to have more features. But sometimes we should balance it with project complexity. This seems a good chance to consolidate a couple things.

maiko3tattun:
> We keep it simple and disable "File", "Edit" and "Project" menus entirely.

Yes, of course we can. Which do you think looks better?
1. image
2. image

> Remove recovery popup and move them to the welcome page. We will also be able to show multiple recoveries this way.

@oxygen-dioxide also mentioned this in #1509. I thought it would be better to wait for his development.

> Get rid of the preference of disabling welcome page. This way we can always show the recoveries.

ok👍️

stakira:
I assume 2 will need to add IsEnabled to every menu item. I think 1 is fine. It will require only one IsEnabled. That's what I mean by keeping it simple.

PR #1551, opened 2025-05-24
Improve loading popup consistency and visuals

Kabinet0:
* Made a lot of loading logic async so the UI can keep rendering while loading

* Fixed broken rendering of the loading popup

* Made loading popup only appear after 100ms of load time to avoid flashing

* Improved aesthetics of loading popup

https://github.com/user-attachments/assets/9a73e1a8-729c-4136-b40b-7073ffeed5a7

https://github.com/user-attachments/assets/d3b9bab4-27f5-4571-96a1-ea4f002b75aa

### DIFF
--- a/OpenUtau/ViewModels/MainWindowViewModel.cs
+++ b/OpenUtau/ViewModels/MainWindowViewModel.cs
@@ -89,18 +89,18 @@ namespace OpenUtau.App.ViewModels {
             RecentFiles.AddRange(Preferences.Default.RecentFiles
                 .Select(file => new RecentFileInfo(file))
                 .OrderByDescending(f => f.LastWriteTime));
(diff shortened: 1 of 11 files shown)


stakira:
You can't through stuff in tasks and expect them to automatically work safely. I'm fine with the menu loadings but let's not touch project loading.

stakira:
Yea I'm aware how the loading popup is having issue, etc. TBH, I'd rather just freeze the UI thread for a moment without any loading window, just to keep things simple. The dotnet tasks are so easy to grow into a spagetti mess.

Kabinet0:
Apologies for the force push, there was some confusion on my part with upstream changes. Everything should be good now.

Though I personally think the extra effort in maintaining an async project loading system is worth it, I respect your decision. I've reverted the project loading changes and removed the popup when opening a project.

PR #1552, opened 2025-05-24
Attack/Release Time of envelope

maiko3tattun:
## New Features:
- Added attack and release times to adjust envelope fade in and fade out

https://github.com/user-attachments/assets/025d0e1f-fc22-4af2-8e49-8f8f4d72ab2d

- ~~Double-clicking on the envelope shows a dialog where you can enter values for the phoneme offset, preutter delta, overlap delta, attack time and release time~~
image

## Fixes:
- Set maximum and minimum values for preutter and overlap commands to prevent envelope collapse
- Adjusted values so that consonants are less likely to be shaved off when overlap is negative (mainly for plosives)

## Note:
- ~~This includes the contents of #1405.~~
- In the future I would like to add an option to manually disable the crossfade.
- Japanese translation memo: 先行発声、オーバーラップ、アタックタイム補正値、リリースタイム補正値


stakira:
Could we keep the behavior of dragging overlap unchanged? I.e., it should control the overlap slope for both the previous and current phoneme, like before. We shouldn't make existing projects suddenly work differently.

maiko3tattun:
The point of overlap has been moved to the lower right corner and is still available. As for the structure of the project, the attack and release to manipulate the fade have only been added; the existing parameters have not been changed.

stakira:
The point is the behavior of projects with modified overlap (the left top point) will change. They will not sound the same as before.

Besides that, we have the side panel. There is no reason to add another UI for them.

Of course, the note properties panel only shows expressions of the first phoneme. That's something we still need to fix.

maiko3tattun:
> The point is the behavior of projects with modified overlap (the left top point) will change. They will not sound the same as before.

Compatibility with existing projects is not destroyed. In the new envelope, the lower left is preutter, the upper left is attack time (default is 0), the upper right is release time (default is 0), and the lower right overlap of the next note. This arrangement is more intuitive. The UI has changed but the envelope is still shaped based on the overlap values.
The ability to change the angle of the fade is very important because Japanese, and especially CV banks, are built on the assumption that there will be no crossfade.

stakira:
I see, the feature is nice but the UX change is kind of disruptive. Let me think about it.

maiko3tattun:
I tried adding a tooltip. Does this solve the problem?

https://github.com/user-attachments/assets/b2418b1c-45f1-4597-8cf3-e9a206b983c9

maiko3tattun:
This PR fixes the rendering error “Index was outside the bounds of the array” because the envelope has been modified.

nagotown:
Could the fixes be split into their own PR? It seems like the new features are still a concern for merging.

stakira (approved):

PR #1553, opened 2025-05-25
Revert "move DLLs to a subfolder"

oxygen-dioxide:
Because netbeauty has a bug that causes OpenUtau unable to launch if it is built with `dotnet publish`. It only works with `dotnet build`.


stakira:
See... I said I'm not too fond of hacky things like this, which you don't know if will break somewhere / someday.

PR #1571, opened 2025-06-04
Use Wine directly instead of wrappers on Mac and Linux

atouu:
Changes:
- Add Wine Path setting in preferences

image

- Use Wine directly when rendering using windows only resamplers and wavtools

https://github.com/user-attachments/assets/80186081-3617-41ca-b538-b8bef53b333e

- Add Support for Legacy Plugins on Mac and Linux

https://github.com/user-attachments/assets/70156715-4d30-4113-bcdd-492d7f5db7bc

- Add Support for Wavtools on Mac

Extra:
- Add Unix Executable in FilePicker file types.

---
Special thanks to @kkgarfin for testing this on Mac.

### DIFF
--- a/OpenUtau.Core/Classic/ExeInstaller.cs
+++ b/OpenUtau.Core/Classic/ExeInstaller.cs
@@ -15,16 +15,6 @@ namespace Classic {
                 : PathManager.Inst.ResamplersPath;
             string destName = Path.Combine(destPath, fileName);
             File.Copy(filePath, destName, true);
-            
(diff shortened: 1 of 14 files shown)


juuyokka:
Confirmed working with moresampler on Arch Linux

odeinjul:
Thanks for your PR! I’ve always found it annoying that users had to write their own scripts just to call Wine to run a resampler, so I was really glad to come across this PR. Looking forward to using the next beta version!

PR #1577, opened 2025-06-11
fix legacy plugin output

takunnma5286:
The initial output for Legacy Plugin was [#PREV], which caused issues with plugins like "おま☆かせ2020" not functioning correctly.
←UTAU OpenUTAU→
image

In accordance with UTAU specifications, I replaced [#PREV] with [#0000] to ensure proper compatibility.
image


takunnma5286:
However, further investigation is needed:

- In UTAU, [PREV] may or may not be present depending on the context.
- Additionally, note numbers can start from an arbitrary index (not always [#0000]).

takunnma5286:
The problem was that it started from [# 0000] even when a [#PREV] value existed.

takunnma5286:
The test is failing as expected due to the discrepancy in the output. However, since altering the test itself may be risky, I will wait for further review from others.

takunnma5286:
To address errors caused by certain legacy plugins that require a fixed 'Envelope' value, it is temporarily output as a fixed value.

nagotown:
This PR prevents VSPitch from working properly
[VSPitch Homepage](https://ackiesound.ifdef.jp/download.html#vspit)

https://github.com/user-attachments/assets/41f05cde-e344-4f38-9ffb-965f3c543bb8

nagotown:
This PR also breaks UtaUtaUtau's [lab2ust](https://github.com/UtaUtaUtau/nnsvslabeling)

PR #1580, opened 2025-06-14
OpenUtau.Core: Add support for /usr/share and /usr/local/share Singers on Linux

dmpmem:
Hey! I'm currently [packaging OpenUtau for my personal Alpine Package Repository](https://codeberg.org/dmpmem/alpine-packaging/src/branch/master/src/utau/openutau) and want to be able to package additional separate packages containing Singers in /usr/share (the system-wide package-manager-managed equivalent to `"${XDG_DATA_HOME:-"$HOME/.local/share"}"`) - as one cannot package things targetting the user's home directory.

This PR adds support for this. It's quite small. I'm not sure what else to write here.

This should probably be extended to most of the other subdirectories of DataPath in future, however this is sufficient for my short-term needs.

### DIFF
--- a/OpenUtau.Core/Util/PathManager.cs
+++ b/OpenUtau.Core/Util/PathManager.cs
@@ -65,11 +65,15 @@ namespace OpenUtau.Core {
         public string RootPath { get; private set; }
         public string DataPath { get; private set; }
         public string CachePath { get; private set; }
+        public string[] SystemWideDataPaths => OS.IsLinux() 
+            ? (Environment.GetEnvironmentVariable("XDG_DATA_DIRS") ?? "/usr/share:/usr/local/share").Split(':').Select(v => Path.Combine(v, "OpenUtau")).ToArray() 
+            : Array.Empty<string>();
         public bool HomePathIsAscii { get; private set; }
         public bool IsInstalled { get; private set; }
         public string SingersPathOld => Path.Combine(DataPath, "Content", "Singers");
         public string SingersPath => Path.Combine(DataPath, "Singers");
         public string AdditionalSingersPath => Preferences.Default.AdditionalSingerPath;
+        public string[] SystemSingersPath => SystemWideDataPaths.Select(path => Path.Combine(path, "Singers")).ToArray();
         public string SingersInstallPath => Preferences.Default.InstallToAdditionalSingersPath
             && !string.IsNullOrEmpty(Preferences.Default.AdditionalSingerPath)
                 ? AdditionalSingersPath
@@ -94,6 +98,11 @@ namespace OpenUtau.Core {
                 }
                 if (Directory.Exists(AdditionalSingersPath)) {
                     list.Add(AdditionalSingersPath);
+                }
+                foreach (string path in SystemSingersPath) {
+                    if (Directory.Exists(path)) {
+                        list.Add(path);
+                    }
                 }
                 return list.Distinct().ToList();
             }


dmpmem:
I'm not sure what equivalent directories on Windows/MacOS are (/if any even exist in the case of MacOS), so I can't really comment on adding those.

atouu:
Where would OpenUtau put resampler generated files (e.g. .llsm, .mrq) if singers from /usr/share/ are read only?

oxygen-dioxide:
Stakira said that we shouldn't have multiple data paths because singers are indexed by their relative path to singers dir. The same singer installed twice to different places will only have one shown up.
Maiko is proposing a custom data path feature: https://github.com/stakira/OpenUtau/pull/1513

dmpmem:
> Where would OpenUtau put resampler generated files (e.g. .llsm, .mrq) if singers from /usr/share/ are read only?

Simply keep those in the user's home directory; these just exist for serving files provided by, say, the distribution.
Unsure if this requires any code changes, but :shrug: it seemed to to the extent I tested it work on this tree, so if I broke anything I didn't notice

> Stakira said that we shouldn't have multiple data paths because singers are indexed by their relative path to singers dir. The same singer installed twice to different places will only have one shown up. Maiko is proposing a custom data path feature: #1513

This is a single custom path; still doesn't solve the desired "hey I want to provide packages for singers and whatnot". If a custom path solution with multiple paths, not all necessarily being writable (see above) is an option, this would work

atouu:
If the singer directory is read only, it'll error with Access Denied since resamplers still generate files to singer's directory and [OU just copies it](https://github.com/stakira/OpenUtau/blob/c3ff44b01f5dcd0ef1bbdb5aa11ef1319cab3a03/OpenUtau.Core/Classic/ClassicRenderer.cs#L69). Unless the packaged singer ship with generated files from all resampler imaginable, most people only includes .frq. That's just how UTAU voicebanks work culturally.

Also, I'm not sure if this applies for system package managers but package manager for singers is listed as won't do in OU's Trello https://trello.com/c/Sf1Djsjs.

oxygen-dioxide:
Do you mean that you want to be able to install singer through Linux's system-wide package manager? I don't think we need this, because a voicebank is pure data, not a software, so they shouldn't be managed through the OS-layer way.

Most voicebank authors don't use Linux. They aren't able to author a Linux package. so even you make this possible, users still have to install voicebanks in the traditional way (with .zip file).

> I'm not sure what equivalent directories on Windows/MacOS are (/if any even exist in the case of MacOS), so I can't really comment on adding those.

For windows and MacOS, there aren't equivalent. Though we have package managers like scoop, they only manage standalone apps and command line tools. They don't manage "datanbase made for another software".

adlez27:
Voicebanks should be treated like media assets rather than software dependencies. I'm also not sure why you need to package voicebanks with an OS. In any case, I'd suggest referring to other sample-based instrument software (like SFZ players) to see how the instrument libraries are handled.

dmpmem:
> with an OS

it's mostly convenience for my personal repository - this isn't me trying to upstream it into upstream alpine :)

life tends to be easier when managing personal setups by packaging them for the distros one uses


yes i'm trying to package singers system-wide, but it's mostly just a "i don't want to need to source things and manually inform the application about each individual singer" - kinda similar in theory to openttd's modular assets [having](https://pkgs.alpinelinux.org/package/edge/community/x86_64/openttd-opensfx) [packages](https://pkgs.alpinelinux.org/package/edge/community/x86_64/openttd-opengfx) despite being non-software assets that can be provided or replaced by selecting the relevant files from the software
these also aren't hard dependencies of openttd; you can just leave them out and run openttd with externally sourced assets

oxygen-dioxide:
Do you use voicebanks made by yourself or use voicebanks made by others? Most voicebanks don't allow redistribution.
You can write a sh script to download and place the voicebanks to OpenUtau's singer folder after installing OpenUtau.

Also, the user won't be able to modify the oto.ini and voice color. I often see voicebanks with oto.ini mistakes and need to manually fix them. In current design, I can just edit the oto.ini inside OpenUtau or with vlabeler. If the voicebank is readonly, I'll have to ask the voicebank author for a bugfix, and wait until they update the voicebank before I continue on my music. 

The examples you listed, [OpenSFX](https://github.com/OpenTTD/OpenSFX) and [OpenGFX](https://github.com/OpenTTD/OpenGFX), are well-maintained opensource projects. Their communities are active and bugs will get fixed quickly. However, most UTAU voicebanks are only managed by one person, and often they may not be enthusiastic on singing voice synthesis anymore and leave the voicebank discontinued. So we can't expect a bugfix from UTAU voicebank authors.

stakira:
Unfortunately, this is a super niche use case that requires a very specific mechanism that doesn't work cross-platform (in fact, not even across Linux distros). It would not make sense to be included in the scope of project.

dmpmem:
> Unfortunately, this is a super niche use case that requires a very specific mechanism that doesn't work cross-platform (in fact, not even across Linux distros). It would not make sense to be included in the scope of project.

(should be compliant with the XDG specification but...) ok!

PR #1592, opened 2025-06-28
Beta won't ask for downgrading to stable

oxygen-dioxide:
Before this PR, if the user downloaded the beta version from GitHub, it will ask the user to "upgrade" to the stable release, which is misleading, especially for diffsinger users who have to use beta version.
After this PR, beta packages downloaded from GitHub will default to beta release channel.


stakira:
Wouldn't this make the beta switch in preference useless?
I think what we need is a default-prefs.jon file that is only used to create new prefs.jon file on first launch. Or we could simply patch Preferences.cs for certain builds.

PR #1595, opened 2025-07-01
Adjustments to DiffSinger phonemizers for indirect multilingual support

lcpdecastro:
Currently, when DiffSinger phonemizers are given phonetic hints inside a note, it follows a two-step process to parse input phonemes into singer-supported phonemes before returning an empty string:

1. Try adding the phoneme as is (ex. `[a]`)
2. Try adding the phoneme prefixed with the phonemizer's language code (ex. `[ja/a]`)

This creates a problem for multilingual voicebanks that *indirectly* support other languages outside of the ones they directly support using `dsdict-xx.yaml` files (ex. using `[ja/a]` as a substitute for `[es/a]`).

To fix this, a third step was added to `DiffSingerBasePhonemizer.ValidatePhoneme()` that tries to substitute phonemes using the voicebank's `dsdict-xx.yaml` file before returning an empty string when a substitute is not found.

*(Note that this issue does not exist when lyrics are input without phonetic hints, as the underlying phonemes are substituted automatically in this case.)*


oxygen-dioxide:
`g2p.Query(phoneme)` seems wrong, because `g2p.Query` means converting plain word into phonemes. Why would you use g2p.Query on a phoneme?

Please provide an example dict file, and screenshots of both old and new behaviours

lcpdecastro:
Here are some screenshots of the Spanish word *contigo* placed twice on the piano roll: first as the word `contigo` and then as the phonetic hint `[k o n t i G o]`. All screenshots were taken in OpenUtau v1.550 with the `DIFFS ES` phonemizer in use.

1. KIT Lite v2.0 (supports Spanish directly)
image

2. Monone v151 (supports Spanish indirectly)
image

3. Tarou Iwata v2 (does not support Spanish)
image

For reference, this is the `replacements` portion of Monone's `dsdict-es.yaml`.
```
replacements:
- {from: a, to: ru/a}
- {from: e, to: ru/e}
- {from: i, to: ru/i}
- {from: o, to: ru/o}
- {from: u, to: ru/u}
- {from: b, to: ru/b}
- {from: B, to: ru/bj}
- {from: ch, to: ru/ch}
- {from: d, to: ru/d}
- {from: D, to: en/dh}
- {from: f, to: en/f}
- {from: g, to: ja/g}
- {from: G, to: ru/gj}
- {from: gn, to: ru/nj}
- {from: I, to: ja/i}
- {from: k, to: ja/k}
- {from: l, to: ru/lj}
- {from: ll, to: ja/y}
- {from: m, to: ja/m}
- {from: n, to: ja/n}
- {from: p, to: ru/p}
- {from: r, to: ru/r}
- {from: rr, to: ru/r}
- {from: s, to: ru/s}
- {from: t, to: ru/t}
- {from: U, to: ru/u}
- {from: w, to: en/w}
- {from: x, to: ru/h}
- {from: y, to: ja/y}
- {from: Y, to: ja/y}
- {from: z, to: ru/s}
- {from: H, to: h}
- {from: J, to: jh}
```

As for the use of `g2p.Query()`, it's been a while since I last contributed to the repo, and my first time contributing to something DiffSinger-related. As such, I wasn't aware that `g2p.Query()` isn't the right way to perform phoneme substitution. I've since found after further testing that since word conversion isn't case-sensitive, phonemes like `[G]` and `[B]` would end up being treated as `[g]` and `[b]`, which wouldn't be correct. Shown below is a screenshot from my *edited* OpenUtau of Albus Korde (who shares the same `dsdict-es.yaml` as Monone above) singing the word `adorable` and the phonetic hints `[a D o r a B l e]` and `[a d o r a b l e]`. The two phonetic hints end up being treated as the same, which shouldn't happen.

image

What would be the correct way of substituting phonemes in this case? I'm not exactly sure where in `g2p` I can retrieve some kind of lookup table for a voicebank's phonemes, if there is a way to do so.

lcpdecastro:
I've made an adjustment to how phoneme substitution is made: instead of using `g2p.Query()`, I've made `ParsePhoneticHint()` build out the replacement list and pass this as a `Dictionary` into `ValidatePhoneme()`. Not exactly sure if this is the most ideal way to do it, since the creation of the `Dictionary` is done for every call to `ParsePhoneticHint()`; maybe this `Dictionary` can be made outside of the method so this isn't done as often. If someone with more experience in C# can improve this, please do (I'm mostly a JavaScript person, so I've had to do a lot of trial and error to make this work).

Here are four voicebanks tested with the new modifications, where inside the notes I've placed the lyrics `contigo`, `[k o n t i G o]` and `k o n t i G o` (without brackets):

1. Kohaku Merry (supports Spanish with `es/` prefixes)
Screenshot 2025-07-02 120915
2. Albus Korde (supports Spanish through substitution)
Screenshot 2025-07-02 120920
3. Tarou Iwata (does not support Spanish)
Screenshot 2025-07-02 120923
4. REO (does not support Spanish, but has unprefixed phonemes. Note the absence of the `[G]` phoneme, which does not appear in his PT-BR phonemes.)
Screenshot 2025-07-02 120929

yqzhishen:
So if I understood correctly, you added a new feature to the DiffSinger dictionary system, that replaces some phonemes to others to _support_ another language?

My biggest doubt is why not use phoneme merging instead. You can train a model with your merged phoneme group, or adjust them by editing phonemes.json in the exported voicebank. I think these can be better than changing the structure of the dictionary file which brings extra costs for maintenance.

lcpdecastro:
> […] you added a new feature to the DiffSinger dictionary system […] I think these can be better than changing the structure of the dictionary file […]

I wouldn’t say I *added* a new feature; it seems that the `dsdict-xx.yaml` files already exist for this purpose. The Lunai banks, in particular, have `dsdict` files for both directly and indirectly supported languages. No restructuring is needed as the `replacements` are already being used when inputting words in notes.

> You can train a model with your merged phoneme group, or adjust them by editing phonemes.json in the exported voicebank.

I personally am just a voicebank user, not a trainer or author. As much as I’d like to retrain a voicebank, it’s not something I plan on doing anytime soon—and I think this applies to most casual OpenUtau users who just want to make covers with already published voicebanks.

The main reason I’m proposing these changes is because I find myself using phonetic input in cases where I want more control of how a word is pronounced. Specifically, I’ve been making the Lunai voicebanks sing in Spanish and Italian, languages they don’t support directly. Currently, phoneme hints in notes are not substituted if they are not provided by the voicebank, prefixed or otherwise. As a result, no phonemes are applied, leaving just the empty `error` phoneme. I’m not exactly keen on having to type, say, `[ja/k en/ao fr/z fr/ah]` when I can just easily type `[k OO z a]` and have `dsdict-it.yaml` dictate which phonemes to substitute instead.

Another use case for this is retuning an already existing USTX to fit another voicebank whose phonemes may not match that of the original. I don’t always want to fiddle around in the Singers panel just to check what phonemes I can use to substitute if a `dsdict` file can do it for me instead. Granted, *I* personally adjust them to improve a voicebank’s pronunciation, but the casual OpenUtau user might not know where to find them (or might not even *want* to edit them).

PR #1597, opened 2025-07-04
Add [FIL VCV & CVVC] and [FIL to JA] phonemizers

Cadlaxa:
both phonemizers doesn't rely in g2p's for now
- [FIL VCV & CVVC] supports phoneme replacements in `filipino.yaml`
- [FIL to JA] same mechanics to [EN to JA]


Cadlaxa:
forgot to mention, can add glottal stops [q] via `'` in the front or end of the note, users can replace the `q` phoneme to other phonemes to fit the vb's format

PR #1618, opened 2025-07-21
Add support for native Linux and MacOS wavtools

atouu:
After #1571, classic renderer no longer parse wavtool wrappers to generate temporary batch scripts; therefore, we can directly call native Linux and MacOS wavtools now.

Demo using wavtool-yawu (wavtool-pl also works):

https://github.com/user-attachments/assets/c73161e6-a214-468c-8869-16a96aedfdeb

Extra:
- Call native Linux/MacOS resamplers directly when using Windows wavtools (similar to worldline).

https://github.com/user-attachments/assets/a39329cb-32e6-471a-aa46-d1786320735a


OpenUtau.Core/Classic/ExeResampler.cs
```diff
@@ -16,6 +16,7 @@ namespace OpenUtau.Classic {
     internal class ExeResampler : IResampler {
         public string Name { get; private set; }
         public string FilePath { get; private set; }
+        public bool CallDirectly { get; private set; }
```
stakira: Isn't this equivalent to not useWine?

atouu: No, since useWine is false in windows. I called it CallDirectly because of WorldlineResampler.

linux-leo:
bumping because I'd love to see this merged.

PR #1620, opened 2025-07-21
Select notes when dragging the keyboard

maiko3tattun:
## New Feature:
Select notes of the same pitch when dragging the keyboard in the pianoroll.
This is similar to the behavior of some DAWs. It is useful when creating harmonies.

https://github.com/user-attachments/assets/0435929d-15eb-477e-ae2f-189b7e329400


maiko3tattun:
Changed to target only the selected note when multiple notes are initially selected.

https://github.com/user-attachments/assets/ec4e19a4-cb71-4f2f-9006-953cab548c1b

stakira:
Most of the time I only want to find a pitch when I drag the keyboard?

maiko3tattun:
When we create harmonies, we often want to select all notes of a certain pitch. A good way to select these without having to drag around the piano roll is to click on the keyboard.

nagotown:
Maybe it could be accessed using a modifier key? If it was <kbd>Shift</kbd>+Drag, it would be consistent with the selection of notes in the piano roll.

maiko3tattun:
I’ve changed it to only select when Shift is pressed. That said, I don’t think there are many situations where accidentally selecting a note by pressing a key would actually be a problem. And I don't think many users would be able to figure this out on their own...

maiko3tattun:
When creating harmonies, I imagine there are many instances where you need to select and correct a group of notes that are out of scale.

PR #1651, opened 2025-08-09
Adds P0 to [- C]s in the Japanese Presamp Phonemizer

AnAndroNerd:
Only works if "CFLAGS" in the Presamp.ini is set to "p0", by default it is.


AnAndroNerd:
Will be updated to use the real value of the "p" flag from the presamp.ini. Edit, i couldn't figure it out

PR #1663, opened 2025-08-18
Remove Spanish (Spain)

AnAndroNerd:
Removes Spanish (Spain), automatically switches the user over to Spanish (Mexico). This is necessary as Spanish (Spain) and Spanish (Mexico) don't have major differences in the context of OU, and Spanish (Spain) has been abandoned and doesn't get updates anymore. Community feedback is needed to see if this is the best way to do this.


AnAndroNerd:
Some other ideas for this are to add the word "deprecated" to Spanish (Spain) for people who would still want to use it. Another idea is to add a way to make es-ES fallback to es-MX instead of en-US.

Nexys4t:
Es una vergüenza el cómo yo propuse esto hace años y sólo movéis el culo cuando queréis tirar a la basura el trabajo que hice en una mañana entera porque no quisisteis hacer nada. Me entristece, me da rabia, nadie ha tenido los huevos de consultarme cuando yo sí tuve las pelotas de preguntaros a vosotros qué hacer.

lottev1991:
@Nexys4t There's no need to be rude in your reply. You could've worded your objections in a polite manner instead of insulting the project and its contributors. I don't think anyone appreciates that attitude here; I sure don't.

AnAndroNerd:
I don't believe that is them being rude, not by Spanish standards at least.

PR #1669, opened 2025-08-23
Add Millefeuille and Marzipan Phonemizers

imsupposedto:
Adding Millfeuille (French) and Marzipan (German) phonemizers and lyrics helpers


imsupposedto:
Error found in German Phonemizer

PR #1674, opened 2025-08-26
Add Commonnote Batch Edits

AnAndroNerd:
Adds Batch edits to support [the commonnote copy and paste standard](https://github.com/ExpressiveLabs/commonnote).

### DIFF
--- a/OpenUtau.Core/Editing/NoteBatchEdits.cs
+++ b/OpenUtau.Core/Editing/NoteBatchEdits.cs
@@ -4,6 +4,10 @@ using System.Linq;
 using System.Threading;
 using OpenUtau.Core;
 using OpenUtau.Core.Ustx;
+using Newtonsoft.Json;
+using static System.Net.Mime.MediaTypeNames;
+using OpenUtau.Classic;
+using TextCopy;
 
 namespace OpenUtau.Core.Editing {
     public class AddTailNote : BatchEdit {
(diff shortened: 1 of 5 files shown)


oxygen-dioxide:
An explaination of what commonnote does:

The clipboard is a fundamental part of desktop productivity. We can seamlessly copy and paste text and images between applications like web browsers and word processors. However, this level of interoperability doesn't exist for musical notes.
Currently, if a user wants to move musical notes between different music creation applications—such as a DAW, MuseScore, or a vocal synthesizer—they have to go through a cumbersome process:
- Manually export a MIDI file from the source application.
- Manually import that same MIDI file into the target application.

This workflow is inefficient and breaks the natural flow of creative work.
To solve this, we are proposing Commonnote, a new, standardized clipboard data format designed specifically for exchanging note data.
With Commonnote, a user could simply copy notes from one Commonnote-supported application and paste them into another, just like they would with text or images. This would significantly streamline the creative process, allowing musicians and producers to move ideas between their favorite tools with unprecedented ease.

AnAndroNerd:
@oxygen-dioxide There seems to be a bug, I'll try my best to fix it.
image

AnAndroNerd:
another bug is there is not an "identifier" when copying and there is no check to see if there's one when pasting. i will try to fix these as well.

AnAndroNerd:
I don't know how to make it check for the identifier when pasting in data, however I doubt that will be much of an issue.

yqzhishen:
A so-called _standard_ with only 3 stars on GitHub is not a real standard. Questions are:
1. How many softwares in the industry already support or are willing to support this standard?
2. Have you done some research on how to design the structure to fit in most softwares?
3. Is it proper for a standard to only define a few basic attributes and leave everything else in the extra map for the implementation?

AnAndroNerd:
Please keep in mind that this is a new standard, and there is nothing else like it at the moment. The team who made it is more than willing to expand it. 

- Two commercial synths plan on using this as their default clipboard, Mikoto Studio and Maghni. To my knowledge, musescore isn't against supporting this. Users are free to make their own implementations to software with scripting support.
- Most midi applications have the same fundamental data, all of which is supported by commonnote.
- It was intentionally designed that way to ensure compatibility. Things that seem as simple as dynamics can be handled wildly differently from software to software.

Keep in mind, I have no involvement in the development of Commonnote. Please direct questions related to it to the GitHub repository.

stakira:
I suspect how useful it is, but the implementation is quite unintrusive. So yea why not.

PR #1687, opened 2025-09-11
Tie values to USTx Resolution

AnAndroNerd:
Replaces values of "480" with the current USTx Resolution to add support for changing midi resolution. This is useful for people who aren't used to the resolution being 480 or people who want more fine control of note size.

### DIFF
--- a/OpenUtau/ViewModels/NotesViewModelHitTest.cs
+++ b/OpenUtau/ViewModels/NotesViewModelHitTest.cs
@@ -57,8 +57,8 @@ namespace OpenUtau.App.ViewModels {
 
     class NotesViewModelHitTest {
         private readonly NotesViewModel viewModel;
-        public ViewConstants ViewConstants = new ViewConstants();
-        public Core.Ustx.UProject project = new UProject();
+        private ViewConstants ViewConstants = new ViewConstants();
+        private UProject project => DocManager.Inst.Project;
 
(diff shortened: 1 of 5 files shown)


AnAndroNerd:
Everything on the UI side is now working

AnAndroNerd:
Outside of importing USTs/USTxs into a project with with a resolution that's not 480, and exporting USTs everything seems to be working.

OpenUtau/ViewModels/NotesViewModelHitTest.cs
```diff
@@ -57,6 +57,8 @@ public struct AliasHitInfo {
 
     class NotesViewModelHitTest {
         private readonly NotesViewModel viewModel;
+        public ViewConstants ViewConstants = new ViewConstants();
+        public Core.Ustx.UProject project = new UProject();
```
AnAndroNerd:
```suggestion
        public UProject project => DocManager.Inst.Project;
```

stakira:
tbh I don't view this as a necessary complexity

AnAndroNerd:
I feel like if you're able to define the resolution in the USTx, you should be able to change it and not have everything break. If not, it raises the question of why you're able to define it in the first place. 
I also think it makes more sense from a reading the code perspective, 480 can seem like a random number for people who aren't familiar with MIDI stuff, but using resolution is easier to understand.

I would also personally use this feature.

PR #1704, opened 2025-09-16
Add Disable Note Size Cap setting

AnAndroNerd:


AnAndroNerd:
Because I made a change that messed with the USTx that could lead to compatibility issues, should we update the USTx version?

stakira:
How is this a problem in practise use? You can barely see such a small note.

AnAndroNerd:
I often need to make notes this small when I'm working with voicebanks that don't have phonemizers.

AnAndroNerd:
Defining a note cap in the USTx seems like a better way of doing this.

PR #1717, opened 2025-09-19
Per singer type required expressions

AnAndroNerd:
Some singer types don't use certain expressions, this allows projects with those singers to remove those expressions.

Known issues

- ~~Crash when opening the expressions menu with no singer~~ this has been fixed

### DIFF
--- a/OpenUtau.Core/Format/USTx.cs
+++ b/OpenUtau.Core/Format/USTx.cs
@@ -7,6 +7,7 @@ using OpenUtau.Classic;
 using OpenUtau.Core.Ustx;
 using OpenUtau.Core.Util;
 using Serilog;
+using ZstdSharp;
 
 namespace OpenUtau.Core.Format {
     public class Ustx {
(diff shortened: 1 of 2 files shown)


OpenUtau.Core/Format/USTx.cs
```diff
@@ -35,7 +36,15 @@ public class Ustx {
         public const string TENC = "tenc";
         public const string VOIC = "voic";
 
-        public static readonly string[] required = { DYN, PITD, CLR, ENG, VEL, VOL, ATK, DEC };
+        public static string[] GetRequiredExpressions(USinger? singer) {
+            if (singer == null) {
```
arx-ein: Since in `ViewModels/ExpressionsViewModel.cs:61` you have
```cs
this.singer = singer ?? USinger.CreateMissing("");
```
to create "missing" singer instead of leaving the `singer` as `null`, and the `CreateMissing()` function is defined as following:
```cs
public static USinger CreateMissing(string name) {
    return new USinger() {
        found = false,
        loaded = false,
        name = name,
    };
}
```
you should check the `singer` value for its `Found` (or `Loaded`, I don't exactly know the difference) property instead of checking if it's `null`, like so:
```suggestion
        public static string[] GetRequiredExpressions(USinger singer) {
            if (singer.Found == false) {
```
This change should solve the "crash when opening the expressions menu with no singer" issue.

stakira:
It would go a bit ugly when a user switch singers. Suddenly the track has every expression.

PR #1720, opened 2025-09-20
Add homebrew to wine search path

atouu:
"Detect" in wine path preference should now attempt to find wine in `/opt/homebrew/bin`

### DIFF
--- a/OpenUtau.Core/Util/OS.cs
+++ b/OpenUtau.Core/Util/OS.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 
 namespace OpenUtau {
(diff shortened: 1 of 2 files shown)


stakira:
Don't you simply want /opt/homebrew/bin in PATH?

atouu:
@stakira  mac user can do that yes, but adding the homebrew path in `/etc/paths` just to detect wine might be complicated for normal mac users. I want to make "Detect" as easy as possible without needing to rely on manually selecting wine.

stakira:
Not /etc/paths, just PATH environment variable. `var values = Environment.GetEnvironmentVariable("PATH");`
Also, for what I remember, homebrew bin is not guaranteed at `/opt/homebrew/bin`. 
Anyway. This doesn't seem a good solution.

atouu:
Homebrew path is only defined in shell not globally. The second statement is true though

PR #1735, opened 2025-09-29
Fix stable release notifications not being sent to beta users

maiko3tattun:


OpenUtau/ViewModels/UpdaterViewModel.cs
```diff
@@ -92,7 +92,7 @@ public UpdaterViewModel() {
                 return null;
             }
             return releases
-                .Where(r => !r.draft && r.prerelease == Preferences.Default.Beta)
+                .Where(r => !r.draft && (!r.prerelease || r.prerelease == Preferences.Default.Beta))
```
arx-ein: Could simplify logic:
```suggestion
                .Where(r => !r.draft && (!r.prerelease || Preferences.Default.Beta))
```

stakira:
Not really necessary though. Stable release may appear to have a +1 version number but should never contain changes not in beta.

maiko3tattun:
Displaying the previous version as the latest can be confusing for users.
image

stakira:
User should receive latest version of the selected channel. It doesn't matter which channel has a higher number.

PR #1746, opened 2025-10-04
Fix voicebank merger error message

maiko3tattun:


stakira:
The majority of the change is reformat. Please don't reformat just for the sake of it.

PR #1747, opened 2025-10-04
Ability to stop rendering and playback

atouu:
Pause vs Stop:
<div align="center">
  App preview
  App preview
</div>
<br>

(Additional) Stop rendering when loading a project:

Stop rendering when loading a project


VocAddict:
As a user I feel like this should be an option in settings or maybe even a switch/knob at in the toolbar/status bar instead of something linked to pause/stop. The thing I like most about OU is that it renders in the background, and I'd hate to make the mistake by clicking one over the other and I lose that ability.

Pause/Stop also have a convention when it comes to DAWs with pause being stop at the current position so it can be resumed from there with stop being to return to last stopped position (unless the settings are changed of course) and it may confuse some users who are used to that.

PR #1749, opened 2025-10-05
Support for install.txt and Plugin Installation

maiko3tattun:
## Development Background
It is common practice to include an install.txt file (http://utau2008.blog47.fc2.com/blog-entry-379.html) in voice bank and plugin ZIP archives.
By consulting install.txt, the installation process should branch based on whether the files should be installed as a voice bank or as a plugin.

## Changes
- Rename `VoicebankInstaller.cs` to `ZipInstaller.cs`
- When install.txt contains `type=editplugin`, it installs as a plugin.
- When install.txt contains `folder=`, sets the plugin installation folder to the specified folder name (voice banks not yet supported).
- When install.txt contains `contentsdir=`, only the specified folder within the zip file is copied (voice banks not yet supported).
- When install.txt contains `description=`, it is displayed during installation.
- A dialog is displayed if the destination folder already exists.

## Translation memo
ファイルの利用規約に同意する場合は次へを押してください。
"{0}" を{1}としてインストールします。
インストール先フォルダ "{0}" が既に存在します。既存のデータとインストールデータは統合されます。

### DIFF
--- a/OpenUtau.Core/Classic/ZipInstaller.cs
+++ b/OpenUtau.Core/Classic/ZipInstaller.cs
@@ -10,7 +10,7 @@ using SharpCompress.Readers;
 
 namespace OpenUtau.Classic {
 
-    public class VoicebankInstaller {
+    public class ZipInstaller {
         const string kCharacterTxt = "character.txt";
         const string kCharacterYaml = "character.yaml";
         const string kInstallTxt = "install.txt";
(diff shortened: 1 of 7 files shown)


stakira:
Please do not mix them up. If a component needs to be shared, it should be made a shared component.

maiko3tattun:
> If a component needs to be shared, it should be made a shared component.

Sorry, could you explain this a bit more specifically?
In the original UTAU, voicebanks and plugins are distributed using a common format. Should I prepare separate views for the voicebank and plugin, extract the zip file when it's dragged and dropped, and then determine which view to pop up?

maiko3tattun:
What should I do if a plugin zip is selected in the Singer installation dialog? Currently it is mistakenly identified as Singer and installed.

rokujyushi:
共通化のためにVoicebankInstaller.csファイルにプラグイン読み込み機能を追加することはNGで
もし共通化するのであれば、ZIPファイル読み込み機能を継承できるベースクラスを作成するか、
または機能を完全に別のソースファイルに分離してほしい、という意味だと私は理解しました。

I thought you meant that adding the plugin loading function to the VoicebankInstaller.cs file for commonality is not acceptable, and if we are to make it common, you would like a base class created that can inherit the ZIP file loading functionality, or you would like the functions separated into a completely different source file.

maiko3tattun:
@rokujyushi ここでいうcomponentってviewのことですよね？
view側はまあsingerとpluginで表示すべきページ数が違うので、今多少複雑になってるので2つ作ってもいいのかなという気もしますが
`VoicebankInstaller.cs`に関してはほぼ同じことしかしてないので分けたところでコード量が2倍になるだけでして…（多少整理する余地はありますが）

PR #1760, opened 2025-10-09
fix: Fix the file path for Linux downloads

creeper-0910:
Fix the file path for Linux downloads


atouu:
Already fixed by ee2ac8888b724d6b99e723111ddddae971a53ba0

PR #1768, opened 2025-10-17
Add Resampler Meta Files definable in resampler manifest

AnAndroNerd:
[moresampler.yaml](https://github.com/user-attachments/files/22961548/moresampler.yaml)
Yaml example


AnAndroNerd:
I'll fix the "Cannot convert null literal to non-nullable reference type." warning, I don't know what the best way to fix it here is.

AnAndroNerd:
Thanks, @Astel123457 for the fix!

Astel123457:
i will note, my fix is possibly not the best way to fix this, but it works.

PR #1774, opened 2025-10-21
feat(onnx): drop Vortice.DXGI & update dml to 1.23.0

fumiama:
> **Note**: This change requires an update on package `Microsoft.ML.OnnxRuntime.DirectML` from `1.16.3` to `1.23.0`.

In some platforms with `Intel(R) AI Boost` NPU, DXGI will report it as a `Microsoft Basic Render Driver` and this will cause `onnxruntime` panic because it actually do not support it.
0bc795fb91103a2c3f1ec1307e38bfa2

In order to eliminate this issue, the native `GetEpDevices` from package `Microsoft.ML.OnnxRuntime` is called, which will replace DXGI and makes it possible to be extended to other device types (like CoreML, OpenVINO, etc.) more easily in the future.

I have tested a DiffSinger model locally and currently no error has been found.

### DIFF
--- a/OpenUtau.Core/Util/Onnx.cs
+++ b/OpenUtau.Core/Util/Onnx.cs
@@ -54,12 +54,12 @@ namespace OpenUtau.Core {
                 var description = "";
                 foreach (var item in device.HardwareDevice.Metadata.Entries) {
                     if (item.Key.ToLower() == "description") {
-                        description = $"[{device.HardwareDevice.Type}] {item.Value}";
+                        description = $"{item.Value} ({device.HardwareDevice.Type})";
                         break;
                     }
                 }
                 if (string.IsNullOrEmpty(description)) { // fallback
-                    description = $"{device.EpName} {device.HardwareDevice.Vendor} {device.HardwareDevice.Type}";
+                    description = $"{device.EpName} {device.HardwareDevice.Vendor} ({device.HardwareDevice.Type})";
                 }
                 devices[i] = device;
                 gpuList.Add(new GpuInfo {


OpenUtau.Core/Util/Onnx.cs
```diff
@@ -35,18 +49,26 @@ public static List<string> getRunnerOptions() {
...
                     }
-                    gpuList.Add(new GpuInfo {
-                        deviceId = deviceId,
-                        description = adapterOut.Description.Description
-                    }) ;
                 }
+                if (description.Length == 0) { // fallback
+                    description = $"{device.EpName} {device.HardwareDevice.Vendor} {device.HardwareDevice.Type}";
+                }
+                devices[i] = device;
```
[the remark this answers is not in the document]
fumiama: Maybe there's some updates of the devices in the system, so it'd better to refresh it when user opens config UI for they possibly want to set some newly installed devices as the EP.

OpenUtau.Core/Util/Onnx.cs
```diff
@@ -66,9 +88,14 @@ private static SessionOptions getOnnxSessionOptions(){
...
             }
-            switch(runner){
+            switch (runner) {
                 case "DirectML":
-                    options.AppendExecutionProvider_DML(Preferences.Default.OnnxGpu);
+                    var d = devices[Preferences.Default.OnnxGpu];
+                    options.AppendExecutionProvider(
+                        OrtEnv.Instance(),
+                        new List<OrtEpDevice> { d } ,
+                        new Dictionary<string, string> {}
```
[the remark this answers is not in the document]
fumiama: It seems that I have already matches your suggestion.

yqzhishen:
What does the ORT API return on your platform after your modification?

fumiama:
> What does the ORT API return on your platform after your modification?

Now it likes this. Although I can filter out CPU, I did not do it since it is a fallback and may benefit someone that panic on `CPU` runner but can run `DirectML` runner without a graphic card.

image

yqzhishen:
> > What does the ORT API return on your platform after your modification?
> 
> Now it likes this. Although I can filter out CPU, I did not do it since it is a fallback and may benefit someone that panic on `CPU` runner but can run `DirectML` runner without a graphic card.
> 
> image

It seems the devices are not on the same indices as those returned by DXGI, But ORT DirectML inference needs the device index. Are you sure this won't cause any problem? Is the inference using the correct device?

yqzhishen:
I read the code and understood your logic. My question is: does `GetEpDevices` return all devices from all EPs? Then, when the user chooses DirectML, can you filter the list so that it contains only devices from DirectML EP?

fumiama:
> I read the code and understood your logic. My question is: does `GetEpDevices` return all devices from all EPs? Then, when the user chooses DirectML, can you filter the list so that it contains only devices from DirectML EP?

Of course I can. The reason why I did not do it has been mentioned above. If you think that it'd be better to filter it out, I can also do it.

fumiama:
After filtering:

image

yqzhishen:
> > I read the code and understood your logic. My question is: does `GetEpDevices` return all devices from all EPs? Then, when the user chooses DirectML, can you filter the list so that it contains only devices from DirectML EP?
> 
> Of course I can. The reason why I did not do it has been mentioned above. If you think that it'd be better to filter it out, I can also do it.

I think filtering the list by EP is better because the EP is chosen from the dropdown above. Also I think the device type `[GPU]` and `[CPU]` is not necessary, because currently most of the EPs only supports a single specific device type.

OpenUtau.Core/Util/Onnx.cs
```diff
@@ -16,6 +15,18 @@ override public string ToString() {
...
 
     public class Onnx {
+        private static readonly Dictionary<int, OrtEpDevice> devices = initializeDevices();
+
+        private static Dictionary<int, OrtEpDevice> initializeDevices() {
+            var env = OrtEnv.Instance();
+            var ortDevices = env.GetEpDevices();
+
+            return ortDevices
+                .Where(device => device.EpName.ToLower().Contains("dml"))
```
yqzhishen, line 25: What do you think if we filter the devices here by user-chosen EP name? If so, the device list can change according to the EP instead of just showing DirectML ones. (Maybe some UI refreshing logic should be implemented too)

fumiama: I think it is OK and we can extend the `runner` selector to filter it and let user add new filters in the list.

However, this is a quite large change. In this case, I would like to keep this PR simple and gradually add the features later when this PR has been proved to be stable.

yqzhishen: OK, maybe I will implement this later in another PR

OpenUtau.Core/Util/Onnx.cs
```diff
@@ -35,18 +46,26 @@ public static List<string> getRunnerOptions() {
...
-                    if(adapterOut is null) {
+            var env = OrtEnv.Instance();
+            var ortDevices = env.GetEpDevices();
+
+            var i = 0;
+            foreach (var device in ortDevices.Where(device => device.EpName.ToLower().Contains("dml"))) {
+                var description = "";
+                foreach (var item in device.HardwareDevice.Metadata.Entries) {
+                    if (item.Key.ToLower() == "description") {
+                        description = $"[{device.HardwareDevice.Type}] {item.Value}";
```
yqzhishen: I think hardware type is not necessary here because currently most of the well-tested EPs only supports a single specific hardware type.

fumiama: I added it here because in the future maybe NPU will also be implemented (through OpenVINO, maybe) and I am working on it. Maybe it can be moved to the tail of the whole string so that user will not get disturbed.

yqzhishen: > I added it here because in the future maybe NPU will also be implemented (through OpenVINO, maybe) and I am working on it. Maybe it can be moved to the tail of the whole string so that user will not get disturbed.

I agree. Using a postfix like `(GPU)` is reasonable.

fumiama: Ok. I will do it tomorrow.

fumiama:
Done.

image

PR #1777, opened 2025-10-25
Add to the list of expressions that lock editing

maiko3tattun:
The project's default expressions must be locked to prevent editing of elements like abbr, type, min, max, etc.
For example, if the abbr is edited, there is a risk that the same expression will be added after applying it, causing a conflict.

Previously, only some expressions were locked.
Now, all default expressions will be added to this list.


stakira:
Why is this a problem?

PR #1779, opened 2025-10-26
Supprt Track Expressions, Import entire flags and Skip flag output when the value is default

maiko3tattun:
### Track Expressions (override)
Override the project expression with the track expression. This is useful when flag ranges or default values differ per resampler.

https://github.com/user-attachments/assets/1d23b656-a1cf-4b44-8b85-5d133b661ee8
### Support Custom Default and Import entire flags（生成時オプション）from ust
Until now, it was unclear whether the default value was the resampler's default specification or the value the user wanted to use.
By introducing the Custom Default feature, you can set the same value for the entire track.
And you can import the original UTAU's entire flags (生成時オプション) as the Custom Default value.

image

### Skip flag output when the value is default
Some resamplers have flags with blank default values. Additionally, when the default flag value differs per resampler, you may want to set the flag value to “none”.
This option is disabled by default. Enabling it suppresses flag output when the phoneme expression matches the default (not a custom default).

## Bug Fix:
- Fixed an issue where all notes had volume 100 and modulation 0 entered when ust was imported, leaving them blank instead.
image

### DIFF
--- a/OpenUtau.Core/Classic/Ust.cs
+++ b/OpenUtau.Core/Classic/Ust.cs
@@ -109,7 +109,7 @@ namespace OpenUtau.Classic {
                         default:
                             if (int.TryParse(header.Substring(2, header.Length - 3), out var noteIndex)) {
                                 var note = project.CreateNote();
-                                ParseNote(note, lastNotePos, lastNoteEnd, block.lines, out var noteTempo, project.expressions);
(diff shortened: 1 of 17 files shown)


maiko3tattun:
- Changed “Custom Default” to “Project Default/Track Default”.
- Applied styles to the expression dialog for easier maintenance.

PR #1804, opened 2025-11-08
Ctrl+Shift+V to paste as plain notes

oxygen-dioxide:
Currently when pasting notes, their expressions, pitch control points and phoneme edits will be pasted. After this PR, user can use Ctrl+Shift+V to paste their notes keeping only positions, durations, tone and lyrics.
图片

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,7 +82,7 @@ jobs:
       - name: DirectML
         shell: cmd
         run: |
-          curl -L https://www.nuget.org/api/v2/package/Microsoft.AI.DirectML/1.12.1 -o Microsoft.AI.DirectML.nupkg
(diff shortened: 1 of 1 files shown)


stakira:
I remember we had a feature to select that to copy and not, maybe they could be merged?

oxygen-dioxide:
> I remember we had a feature to select that to copy and not, maybe they could be merged?

It is "Select and paste parameters", which means pasting the parameters of notes in clipboard onto the existing selected notes. Basically, what we have is "paste parameters only". what we don't have (and I'm adding) is "paste notes only".

OpenUtau已有的功能是仅粘贴参数，不粘贴音符，适用于音符已经有了的情况。我新增的功能是仅粘贴音符，不粘贴参数。

图片

PR #1811, opened 2025-11-09
Disable the menu when undo is not possible / Display the command name

maiko3tattun:
## New Feature:
- Added command names to the undo menu
  - ~~Note that this currently only supports some commands and translation is not yet supported.
I'd like to confirm with you whether I should proceed with this direction for now.~~

## Behavior Change:
- When undo or redo is not possible, the menu has been disabled

https://github.com/user-attachments/assets/7f0afd3e-f63a-4247-b67e-aaa4225ed1c2


stakira:
I think this is good

maiko3tattun:
Ready to merge :)

stakira:
conflict

PR #1840, opened 2025-11-23
DiffSinger Refined Phonemizer Implementation

overdramatic:
# DiffSinger Refined Phonemizer Implementation

## Overview
This PR introduces the `DiffSingerRefinedPhonemizer` - an abstract base class based on `DiffSingerBasePhonemizer` and `DiffSingerG2pPhonemizer`, that provides comprehensive infrastructure for DiffSinger-based phonemizers in OpenUtau. The implementation delivers advanced phoneme processing capabilities with sophisticated phoneme replacement systems based on replacements lists, word-level, timing-based and duration-dependent modifications

## Key Features
- **Advanced Phoneme Processing**: Supports single, merge (M:1), split (1:M), and many-to-many phoneme replacements
- **Multi-language Support**: Handles language-specific phoneme mappings and prefixes
- **Word-Level Editing**: Enables cross-phoneme modifications across entire words
- **Comprehensive G2P System**: Integrates grapheme-to-phoneme conversion with customizable dictionaries

## Benefits
- **Production Ready**: Comprehensive error handling and logging
- **High Performance**: Optimized processing with caching
- **Flexible**: Extensive customization points
- **Robust**: Graceful fallbacks and validation
- **Maintainable**: Clean architecture with clear separation of concerns

This implementation provides a solid foundation for sophisticated singing synthesis with DiffSinger, supporting complex linguistic phenomena while maintaining performance and extensibility.

## Usage
Derive from `DiffSingerRefinedPhonemizer` and override:
- `EditPhonemesForWord()` for word-level modifications
- `EditTimedPhonemes()` for timing-based transformations
- `ApplyDurationBasedReplacements()` for duration-dependent changes

For `dsdict.yaml` replacements:
```yaml
replacements:
- {from: ng, to: pt/g} #Single Replacement
- {from [k, w] to: pt/kw} #Merge
- {from gw to: [pt/g, pt/w]} #Split
- {from: [k, w, an], to: [pt/s, pt/t, pt/r, pt/a]} #Many-to-Many
```
It uses the same rules as `DiffSingerG2pPhonemizer` with `from` being without the language tag and `to` with the tag if the model has multi-dict support
The phonemizer folows the replacements in logical order: single -> merge -> split -> many-to-many

### DIFF
--- a/OpenUtau.Core/DiffSinger/Phonemizers/DiffSingerRefinedPhonemizer.cs
+++ b/OpenUtau.Core/DiffSinger/Phonemizers/DiffSingerRefinedPhonemizer.cs
@@ -1,0 +1,1274 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using K4os.Hash.xxHash;
+using Microsoft.ML.OnnxRuntime;
+using Microsoft.ML.OnnxRuntime.Tensors;
+using OpenUtau.Api;
+using OpenUtau.Core.G2p;
+using OpenUtau.Core.Ustx;
+using OpenUtau.Core.Util;
+using Serilog;
+
+namespace OpenUtau.Core.DiffSinger {
+    #region Data Structures
+
+    /// <summary>
+    /// Dictionary data structure for DiffSinger G2P (Grapheme-to-Phoneme) conversions
+    /// with support for various replacement types (single, merge, split, many-to-many)
+    /// </summary>
+    class DiffSingerRefG2pDictionaryData : G2pDictionaryData {
+        public struct Replacement {
+            public object from { get; set; }
+            public object to { get; set; }
+
+            /// <summary>
+            /// Gets the 'from' values as a list of strings
+            /// </summary>
+            public List<string> FromList {
+                get {
+                    if (from is string s) return new List<string> { s };
+                    if (from is IEnumerable<object> list) return list.Select(x => x.ToString()).ToList();
+                    return new List<string>();
+                }
+            }
+
+            /// <summary>
+            /// Gets the 'to' values as a list of strings
+            /// </summary>
+            public List<string> ToList {
+                get {
+                    if (to is string s) return new List<string> { s };
+                    if (to is IEnumerable<object> list) return list.Select(x => x.ToString()).ToList();
+                    return new List<string>();
+                }
+            }
+        }
+        
+        public Replacement[]? replacements;
+
+        /// <summary>
+        /// Creates a dictionary for one-to-one replacements (single symbol mapping)
+        /// </summary>
+        public Dictionary<string, string> replacementsDict() {
+            var dict = new Dictionary<string, string>();
+            if (replacements != null) {
+                foreach (var r in replacements) {
+                    if (r.from is string fromStr && r.to is string toStr) {
+                        dict[fromStr] = toStr;
+                    }
+                }
+            }
+            return dict;
+        }
+
+        /// <summary>
+        /// Creates a dictionary for merge replacements (many-to-one mapping)
+        /// </summary>
+        public Dictionary<string, string> mergeDict() {
+            var dict = new Dictionary<string, string>();
+            if (replacements != null) {
+                foreach (var r in replacements) {
+                    var fromList = r.FromList;
+                    var toList = r.ToList;
+                    
+                    // If 'from' is an array (more than one element), add to merge dict
+                    if (fromList.Count > 1 && toList.Count == 1) {
+                        string fromKey = string.Join("|", fromList);
+                        dict[fromKey] = r.to.ToString();
+                    }
+                }
+            }
+            return dict;
+        }
+
+        /// <summary>
+        /// Creates a dictionary for split replacements (one-to-many mapping)
+        /// </summary>
+        public Dictionary<string, string> splitDict() {
+            var dict = new Dictionary<string, string>();
+            if (replacements != null) {
+                foreach (var r in replacements) {
+                    var fromList = r.FromList;
+                    var toList = r.ToList;
+                    
+                    // If 'to' is an array (more than one element), add to split dict
+                    if (fromList.Count == 1 && toList.Count > 1) {
+                        dict[r.from.ToString()] = string.Join("|", toList);
+                    }
+                }
+            }
+            return dict;
+        }
+
+        /// <summary>
+        /// Creates a dictionary for many-to-many replacements (array-to-array mapping)
+        /// </summary>
+        public Dictionary<string, string> manyToManyDict() {
+            var dict = new Dictionary<string, string>();
+            if (replacements != null) {
+                foreach (var r in replacements) {
+                    var fromList = r.FromList;
+                    var toList = r.ToList;
+                    
+                    // If both 'from' and 'to' are arrays with more than one element
+                    if (fromList.Count > 1 && toList.Count > 1) {
+                        string fromKey = string.Join("|", fromList);
+                        dict[fromKey] = string.Join("|", toList);
+                    }
+                }
+            }
+            return dict;
+        }
+    }
+
+    /// <summary>
+    /// Represents a phoneme with its symbol and speaker information
+    /// Made public to allow external access to phoneme data
+    /// </summary>
+    public struct dsRefPhoneme {
+        public string Symbol;
+        public string Speaker;
+
+        public dsRefPhoneme(string symbol, string speaker) {
+            Symbol = symbol;
+            Speaker = speaker;
+        }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


hecko-yes:
Might it make more sense to reverse the order of replacements, or even use the order as listed in the .yaml file for maximum flexibility? For example, this isn't possible if you process single replacements before everything else:
```yaml
replacements:
- {from: [s, p], to: [s, p_]}
- {from: p, to: p_h}
```

overdramatic:
> Might it make more sense to reverse the order of replacements, or even use the order as listed in the .yaml file for maximum flexibility? For example, this isn't possible if you process single replacements before everything else:
> 
> ```yaml
> replacements:
> - {from: [s, p], to: [s, p_]}
> - {from: p, to: p_h}
> ```
I don't get what you are trying to say, but this base class follows the replacement order:
`single -> merge -> split -> many-to-many`

For more sophisticated or in depth customization, it's better to make inside the phonemizer code

yqzhishen:
What's the use case of this?

overdramatic:
> What's the use case of this?

@yqzhishen  you can see it in this PR: https://github.com/openutau/OpenUtau/pull/1841

yqzhishen:
@overdramatic I don't think these hardcoded linguistic rules are the right philosophy for a machine learning model, which should have context capability.

overdramatic:
@yqzhishen  can you please tell me what hardcoded linguistic rules do you think it should not be included?
On BRAPA (Brazilian Portuguese) cases, those hardcoded linguistics on our phonemizer, like the word boundary rules, is because the G2P on OpenUTAU is very limited. They work in a word level and not in a sentence level, so it doesn't have context capability to change that said boundaries.
The same goes for the duration based rules: Since a G2P doesn't get those phoneme timings, and, on our tests with DiffSinger, trying to merge `/a/ /i/ /u/` with their reduction variants `/ax/ /i0/ /u0/`, could give inconsistencies on inference, it was our best approach to get those reductions right in the phonemizer.
And those options can be turned off/skipped  if you add a `/` on the start of the word or input the phonemes with `[ ]`

Hope this wikipedia page helps you understand why those hardcoded rules is needed:
[Sandhi](https://en.wikipedia.org/wiki/Sandhi) / [Sandhi Portuguese examples](https://en.wikipedia.org/wiki/Sandhi#Portuguese)

agentasteriski:
Please reconsider the decision to close this. It may not seem helpful to languages with simple phonetic rules where everything can be matched precisely to the writing, but the rest of us have been begging for this to merge.

yqzhishen:
The problem wasn't in G2P; the real problem is that you expected G2P to handle what it shouldn't be responsible for.

You proposed 4 categories of linguistic rules:

1. **duration-driven reduction variants**. DiffSinger is nn-based model, which already takes duration and phoneme IDs as inputs. If the voicebank is made well, it should have already learned to weaken pronunciation on short phones. Adding such manual rules is not only redundant, but also harmful to the semantics of context sequence; not to mention the weird hardcoded threshold which is definitely ungeneralizable.
2. **Sandhi rules across words**. Again, the model receives full sequence, so it already sees these conjunction patterns. These manual rules are naive statistical trends of connected speech phenomena, and won't be better than what the model should have learned.
3. **Dummy or placeholder symbols**. These things blurred the lines of responsibility; current protocol is fragile, not commonly recognized, and hard to maintain. From G2P side, it'd better produce a standard form of a specific phoneme; from voicebank side, the interpretion rules are static, thus won't be better than model predictions.
4. **Singing technique tags**. Out of scope, not G2P-related. `cl` is Japanese sokuon; `vf` isn't a phoneme from phonetics perspective.

The PR mainly follows the philosophy of symbolic phonology, proposing explicit knowledge, cascaded inference and universal rules, where correctness is defined by linguists. However, neural acoustics proposes implicit knowledge and end-to-end mapping, where information and context are fundamental concerns, and the behavior is driven by data.

I would like to quote Frederick Jelinek, pioneer of modern speech recognition and natural language processing research, for his famous statement,

> Every time I fire a linguist, the performance of the speech recognizer goes up.

I hope my explanation help you find the right way to deal with modern singing voice synthesis.

nagotown:
@yqzhishen maybe i'm misunderstanding, but based on the logic being mentioned, shouldn't only the BRAPA phonemizer be denied instead of the API? what if i wanted to set replacements for `tr` -> `t`, `r` on an english phonemizer?

yqzhishen:
> @yqzhishen maybe i'm misunderstanding, but based on the logic being mentioned, shouldn't only the BRAPA phonemizer be denied instead of the API? what if i wanted to set replacements for `tr` -> `t`, `r` on an english phonemizer?

Makes sense, sorry. I will review this when I have time.

If there are further discussion around BRAPA, let's move to #1841.

gnloop:
> The problem wasn't in G2P; the real problem is that you expected G2P to handle what it shouldn't be responsible for.
> 
> You proposed 4 categories of linguistic rules:
> 
> 1. **duration-driven reduction variants**. DiffSinger is nn-based model, which already takes duration and phoneme IDs as inputs. If the voicebank is made well, it should have already learned to weaken pronunciation on short phones. Adding such manual rules is not only redundant, but also harmful to the semantics of context sequence; not to mention the weird hardcoded threshold which is definitely ungeneralizable.
> 2. **Sandhi rules across words**. Again, the model receives full sequence, so it already sees these conjunction patterns. These manual rules are naive statistical trends of connected speech phenomena, and won't be better than what the model should have learned.
> 3. **Dummy or placeholder symbols**. These things blurred the lines of responsibility; current protocol is fragile, not commonly recognized, and hard to maintain. From G2P side, it'd better produce a standard form of a specific phoneme; from voicebank side, the interpretion rules are static, thus won't be better than model predictions.
> 4. **Singing technique tags**. Out of scope, not G2P-related. `cl` is Japanese sokuon; `vf` isn't a phoneme from phonetics perspective.
> 
> The PR mainly follows the philosophy of symbolic phonology, proposing explicit knowledge, cascaded inference and universal rules, where correctness is defined by linguists. However, neural acoustics proposes implicit knowledge and end-to-end mapping, where information and context are fundamental concerns, and the behavior is driven by data.
> 
> I would like to quote Frederick Jelinek, pioneer of modern speech recognition and natural language processing research, for his famous statement,
> 
> > Every time I fire a linguist, the performance of the speech recognizer goes up.
> 
> I hope my explanation help you find the right way to deal with modern singing voice synthesis.

This isn't entirely correct.

I can somewhat agree with point 1.

But something like labeling 's' and 'sh' both as 's' just because "the model will learn the difference based on context" feels very silly. 

When a simple edit in the phonemizer can just turn the 's' into 'sh' BEFORE inference why bother? It's an extremely easy fix and it can be used for multiple languages and multiple use cases, not just BRAPA, and it takes away the guess work for the model (especially when using lower amounts of data).

It just gives people more option without having to make an entirely new phonemizer. 

For example, the new update to the Italian phonemizer broke some things with the previous system, so I'd have to correct all the labels, or I could just use this replacement system (personally I don't agree with the changes, I would prefer the old labeling style).
It also applies to the DIFFS EN+ phonemizer: it forces the user to use tr and dr phoenmes, but not everyone uses those, so they are stuck with the DIFFS EN phonemizer which has less accuracy for general singing purposes.

Going back to BRAPA specifically, letting the AI guess things is a bad idea when taking into account that there are different rules in different Brazilian regional dialects (or european Portuguese) that often contradict each other, so this is the best idea, since the alternative would be making many different phonemizers.

overdramatic:
> > @yqzhishen maybe i'm misunderstanding, but based on the logic being mentioned, shouldn't only the BRAPA phonemizer be denied instead of the API? what if i wanted to set replacements for `tr` -> `t`, `r` on an english phonemizer?
> 
> Makes sense, sorry. I will review this when I have time.
> 
> If there are further discussion around BRAPA, let's move to #1841.

Thank you!
I've made a reply, hope you can see it soon
https://github.com/openutau/OpenUtau/pull/1841#issuecomment-4996547468

AnAndroNerd (approved):

nagotown:
hiding language prefix seems not to work right now (pr made before that feature was made?)
cadlaxa's diffs en++ (using refined api)
OpenUtau_FoPWOOhvES
diffs en+ (also by cadlaxa but built into OU)
OpenUtau_iTHaNXG11V

overdramatic:
> hiding language prefix seems not to work right now (pr made before that feature was made?) cadlaxa's diffs en++ (using refined api) OpenUtau_FoPWOOhvES diffs en+ (also by cadlaxa but built into OU) OpenUtau_iTHaNXG11V

@nagotown it should be working with this new fix

nagotown:
this would also be helpful for japanese multiple replacements for stuff like `っ` -> `GS` when surrounded by vowels

PR #1852, opened 2025-11-30
Fix Theme Switching

atouu:
Fixes a bug where switching from Custom YAML to Light or Dark mode would cause colors to display incorrectly.


atouu:
Already merged with #1888

PR #1859, opened 2025-12-01
Optimizing Window Size Saving

maiko3tattun:
- This PR restores the previous full-screen state while also remembering the size when not maximized.
  - If you quit in full-screen mode last time, ou will launch in full-screen mode again next time, but when you exit full-screen mode, it will return to its original size.
- The welcome screen won't break no matter what size it launches at.

### DIFF
--- a/OpenUtau.Core/Util/WindowSize.cs
+++ b/OpenUtau.Core/Util/WindowSize.cs
@@ -13,11 +13,26 @@
         }
 
         public void Set(double width, double height, int posX, int posY, int state) {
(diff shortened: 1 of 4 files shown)


maiko3tattun:
- If the window was minimized last time, it will launch in normal mode
- If the window was in full-screen (with the taskbar hidden) last time, it will now launch in maximized mode (with the taskbar visible)
- Fixed an issue where LayoutSplit did not work properly on high-resolution monitors in Windows

PR #1861, opened 2025-12-02
feat(onnx): only show gpu selection on dml runner

fumiama:
Before this modification, the GPU selection shows up in every runners, which makes it confusing since the GPU selection result is only used in DirectML runner. After this, the relationship becomes clear, as shown in the screen record below:

https://github.com/user-attachments/assets/dea3d7c0-7371-458a-9b37-bdc3007433c9


fumiama:
Hello @stakira please consider merge this PR since it is quite a small change so that I can do further improvements based on it and avoiding possible code conflict. Please also note that the **macos-13, osx-x64** is retired and canceled, so that it cannot pass all tests.

fumiama:
Hello @stakira thanks for your merging. I have updated this PR and resolved its conflict with main branch. Now it passed all the tests. By the way, the required tests in the PR check is out-of-date so it will pending forever.

PR #1863, opened 2025-12-02
Fix importing USTs

AnAndroNerd:
USTs with flags set in the settings break ever since #1805, this is more of a temp fix than a real one.


maiko3tattun:
Thank you for your fix. Using this as a reference, I've created a further enhanced PR.

PR #1865, opened 2025-12-05
[ENG VCCV] Add support for vcVowels

AnAndroNerd:


OpenUtau.Plugin.Builtin/Data/envccv.template.yaml
```diff
@@ -2,7 +2,7 @@
 ---
 version: 1.1
```
AnAndroNerd:
```suggestion
version: 1.2
```

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -23,7 +23,7 @@ namespace OpenUtau.Plugin.Builtin {
     // cadlaxa here ^_^
     public class EnglishVCCVPhonemizer : SyllableBasedPhonemizer {
         private const string LatestVersion = "1.1";
```
AnAndroNerd:
```suggestion
        private const string LatestVersion = "1.2";
```

AnAndroNerd (commented):
Bump yaml template version

PR #1821, opened 2025-11-14
Custom Theme Support

AnAndroNerd:
Adds support for defining a custom theme in a YAML file.

- [x] "Bind" Colors in axaml from C#
- [x] Change Theme in preferences menu
- [x] Read Colors from an external YAML

Closes #1795.

PR #1867, opened 2025-12-05
Fix colors remaining unchanged after custom theme

arx-ein:
## Overview
Fixed an issue regarding the custom theme feature implemented in https://github.com/stakira/OpenUtau/pull/1821
After a custom theme is loaded, some colors will remain unchanged while switching back to other themes.

## Details
Here, I first loaded the default "Custom YAML" theme (which is basically identical to the light theme).

fix-custom-theme-remaining-00

Then, when I switch back to a different theme (I used dark theme to make it more obvious), some elements like the title bar or texts change color properly but others like the background color remains the same.

fix-custom-theme-remaining-01

With this PR this will no longer be the case, and all the colors will change accordingly as determined by the themes.


arx-ein:
The fix now works fine but is more of a temporary one, so consider this WIP as I will work on a better solution.

arx-ein:
I've been informed that this PR tackles the same issue as https://github.com/stakira/OpenUtau/pull/1852, making this a duplicate, while the changes are different.
I plan to finish my changes nonetheless and compare these afterwards.

arx-ein:
Ended up rewriting `CustomTheme` altogether but I believe now the code is more concise and safer. I'd mark it ready for review.

arx-ein:
Issue resolved with https://github.com/stakira/OpenUtau/pull/1888. Closing the PR.

PR #1858, opened 2025-12-01
Support for Multiple Custom Themes

atouu:
Add support for multple custom themes by placing yaml files in `OpenUtau/Themes` directory.

multi-custom-theme

Technical Changes:
- Replace `Theme` with `ThemeName` in preferences. Current user theme will reset to Light.
- `Light` and `Dark` strings are no longer translatable (unless other contributors can make it work).

This PR depends on PR #1852, either merge that first or only merge this PR.

PR #1869, opened 2025-12-06
Fix for OU not starting when theme.yaml is corrupted

maiko3tattun:
When theme.yaml is corrupted,
- If loading fails during OU startup, an error log is recorded.
- If the theme is changed from settings, an error dialog is displayed.

If this issue will be fixed in #1858, there is no need to merge this PR.


stakira:
Conflict after #1888. Is it still necessary?

AnAndroNerd:
No, it is not still needed.

PR #1876, opened 2025-12-15
Update macOS x64 runner version in github actions workflow

oxygen-dioxide:
According to [github official announcement](https://github.com/actions/runner-images/issues/13046), `macos-13` runner is retired. Macos x64 users should use `macos-15-intel` instead.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,8 +33,8 @@ jobs:
           - { name: win-x64, rid: win-x64, arch: x64, os: win, runs-on: windows-latest }
           - { name: win-x86, rid: win-x86, arch: x86, os: win, runs-on: windows-latest }
           - { name: win-arm64, rid: win-arm64, arch: arm64, os: win, runs-on: windows-latest }
-          - { name: osx-x64, rid: osx-x64, arch: x64, os: osx, runs-on: macos-13 }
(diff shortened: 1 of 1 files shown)


JMii63:
won’t this affect “legacy” MacOS support?

.github/workflows/build.yml
```diff
@@ -33,8 +33,8 @@ jobs:
           - { name: win-x64, rid: win-x64, arch: x64, os: win, runs-on: windows-latest }
           - { name: win-x86, rid: win-x86, arch: x86, os: win, runs-on: windows-latest }
           - { name: win-arm64, rid: win-arm64, arch: arm64, os: win, runs-on: windows-latest }
-          - { name: osx-x64, rid: osx-x64, arch: x64, os: osx, runs-on: macos-13 }
-          - { name: osx-arm64, rid: osx-arm64, arch: arm64, os: osx, runs-on: macos-13 }
+          - { name: osx-x64, rid: osx-x64, arch: x64, os: osx, runs-on: macos-15-intel }
+          - { name: osx-arm64, rid: osx-arm64, arch: arm64, os: osx, runs-on: macos-15-intel }
```
stakira, line 37: Is this line by mistake? arm64 build doesn't need a intel machine

stakira: Actually, I don't think the intel build relies on a intel machine either. macos-14 could probably work for all cases.

PR #1885, opened 2025-12-21
Fix import 24bit wav

maiko3tattun:
Fixed an issue where 24-bit WAV instruments could not be imported.
~~(It would be better if we could verify the behavior on macOS before merging)~~
For now, I've only fixed the Windows behavior. A better fix is expected.


stakira:
I'd rather it be fixed in the NAudio project.

maiko3tattun:
I tried every NAudio read function, but couldn't load 24-bit WAV files. Of course, if it can be fixed with NAudio alone, that would be preferable.

PR #776, opened 2023-07-14
ZH CVV: custom dictionary support

oxygen-dioxide:
- Added a monophone base phonemizer (or C+V phonemizer).
- Migrated ZH CVV to MonophonePhonemizer to support custom dictionary (zhcvv.yaml).

PR #1906, opened 2026-01-05
feat: Custom Dictionary for Cantonese phonemizer

Aldof101:
This PR adds custom dictionary functionality similar to the Chinese CVV (October) phonemizer. The phonemizer now loads YAML dictionary files (zhyue.yaml) from singer and plugin folders, allowing users to define custom phoneme splits while maintaining backward compatibility with built-in rules.

### DIFF
--- a/OpenUtau.Plugin.Builtin/CantoneseSyoPhonemizer.cs
+++ b/OpenUtau.Plugin.Builtin/CantoneseSyoPhonemizer.cs
@@ -1,59 +1,115 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Linq;
+using System.IO;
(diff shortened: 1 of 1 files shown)


Aldof101:
I've included an example custom `zhyue.yaml` file, formatted similarly to the zhcvv.yaml file of the Chinese CVV phonemizer (refer to https://github.com/stakira/OpenUtau/pull/776). Since the Cantonese Syo phonemizer retains the interface for automatically mapping Chinese characters to Jyutping, users can now directly input Chinese characters and they will be automatically converted to Jyutping and then split according to the custom dictionary rules.
[zhyue.yaml](https://github.com/user-attachments/files/24452241/zhyue.yaml)

PR #1907, opened 2026-01-05
Windows x86 Fix and Changes

atouu:
This PR fixes null reference bug by using `msbuild` instead of `dotnet publish`. Since ONNX is not supported in Windows x86 anymore, I also added these following changes:

- Tell users that x86 builds no longer support features that uses Onnx.
   Screenshot_2026-01-05_17-09-32
- Remove DiffSinger, ENUNU and Vogen voicebanks from voicebank list and G2P Phonemizers in phonemizer list.


Cadlaxa:
Is x86 builds still relevant? like most of the features will not work especially with onnx runtime 😔

atouu:
@Cadlaxa  Tbh, I prefer to drop x86 too atp

Cadlaxa:
😭😭

PR #1909, opened 2026-01-06
Fix NullReferenceException when importing UST with unregistered flags

qwqAutomaton:
## Summary
This PR fixes a `System.NullReferenceException` that occurs when parsing UST flags that are not registered in the project's expressions list.

## The Bug
When importing UST with flag into a new UProject, the parser identifies flags correctly, but the `FirstOrDefault` always returns `null` since no expressions are registered to the project.

## The Fix
Added a null check before cloning the expression descriptor. If an expression/flag is registered, it is now safely ignored.

## Future Improvements / Suggestions
Since I am not very familiar with C#, I only fixed the crash with minimum modification. However, silently ignoring unknown flags is not ideal. A better long-term solution would be to implement a Flag Mapping Popup . When importing a UST with unknown flags, the user could be prompted to manually map them to valid/registered USTX flags.

### DIFF
--- a/OpenUtau.Core/Classic/Ust.cs
+++ b/OpenUtau.Core/Classic/Ust.cs
@@ -175,9 +175,14 @@ namespace OpenUtau.Classic {
                         var parser = new UstFlagParser();
                         var track = project.tracks[0];
                         foreach (var flag in parser.Parse(parts[1].Trim())) {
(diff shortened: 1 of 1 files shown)


AnAndroNerd:
This was already fixed in https://github.com/stakira/OpenUtau/pull/1864

PR #1944, opened 2026-01-28
Loading external batch edits

AnAndroNerd:
Example plug-in [here](https://files.catbox.moe/zo5qzm.dll), it turns the lyrics into the word test.

You set up external batch edits in code exactly as you would with normal ones but in a different C# project.

### DIFF
--- a/OpenUtau/Controls/PianoRoll.axaml.cs
+++ b/OpenUtau/Controls/PianoRoll.axaml.cs
@@ -145,35 +145,21 @@ namespace OpenUtau.App.Controls {
                 Command = noteBatchEditCommand,
                 CommandParameter = edit,
             }));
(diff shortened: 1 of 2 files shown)


OpenUtau/Controls/PianoRoll.axaml.cs
```diff
@@ -143,6 +145,35 @@ await MessageBox.ShowProcessing(RootWindow, $"{name} - ? / ?",
                 Command = noteBatchEditCommand,
                 CommandParameter = edit,
             }));
+            try
+            {
+                var pluginsPath = PathManager.Inst.PluginsPath;
```
stakira: plugin search should happen together with other plugins

PR #1966, opened 2026-02-14
Upgrade to .NET 10

AnAndroNerd:
.NET 8 reaches EOL at November 10, 2026, so this PR brings it up. It also makes OU slightly faster. Not necessary right now, but still probably a good idea to merge.


fempolyhedron:
uhm you forgot to change the .net versions for windows in OpenUtau.Test.csproj

PR #1973, opened 2026-02-16
Discord RPC

fempolyhedron:
Added discord rich presence also lists the names of the singers being currently used


fempolyhedron:
So it turns out there is a high chance that it wont work on other oses, it would compile, just the rpc wont show

ghost:
pretty good chance this will only work out-of-the-box on windows and maybe macOS, not sure what linux would look like

atouu:
Not working on linux, probably won't work on macOS too

AnAndroNerd:
I can test on MacOS, but I have to ask, why not use a simple NuGet library for this?

fempolyhedron:
i may have uhm, forgotten, that that exists

OpenUtau/Integrations/DiscordRPC.cs
AnAndroNerd, line 1: I'd put this file in OpenUtau/Integrations.

OpenUtau/Integrations/DiscordRpcSubscriber.cs
AnAndroNerd, line 1: This one as well.

PR #1983, opened 2026-02-23
German CV phonemizer to German

LilyCodes303:
Added German CV to phonemizer list


lottev1991:
This phonemizer should be put into the German section, not General. I suggest that you fix this, otherwise it might get rejected. Just a warning.

Also, I recommend using the Syllable-based Phonemizer API, as well as implementing one of the German G2P models into this phonemizer. That way, people can easily type German lyrics onto notes, and they'll be automatically converted to phonemes. Standalone consonants will also be placed automatically.

LilyCodes303:
> This phonemizer should be put into the German section, not General. I suggest that you fix this, otherwise it might get rejected. Just a warning.
> 
> Also, I recommend using the Syllable-based Phonemizer API, as well as implementing one of the German G2P models into this phonemizer. That way, people can easily type German lyrics onto notes, and they'll be automatically converted to phonemes. Standalone consonants will also be placed automatically.

Ok so. I did add G2P or a very simple version at least... Its just that the phonemizer expects a very specific reclist! I think... And I will fix it being in the wrong catagory today

PR #2004, opened 2026-03-12
[Transcribe] RMVPE support

emeraldsingers:
This PR adds optional RMVPE-based f0 prediction to the audio transcription flow.

After note transcription is completed with either `GAME` or `SOME`, OpenUtau can now run `RMVPE` on the same audio and apply the predicted f0 as a `PITD` curve to the generated voice part.

## Changes

- add `OpenUtau.Core/Analysis/Rmvpe.cs`
- add RMVPE checkbox to the transcription dialog
- apply predicted f0 to the created part as `PITD`
- add UI strings and RMVPE error message

## Behavior

- if `Predict f0 (PITD) using RMVPE` is enabled, RMVPE runs afterward
- the resulting pitch is written into the generated voice part as a `PITD` curve

## Notes

This keeps RMVPE as an optional post-process and does not change the existing note transcription behavior.

## RMVPE Oudep
[GitHub Releases](https://github.com/emeraldsingers/UtauV_Packages/releases/download/rmvpe/rmvpe.oudep)
[Google Drive](https://drive.google.com/file/d/1oyvtbRXmP3IIEXhBefE-LpOr3oekA_ri/view?usp=sharing)


yqzhishen:
I don't think computing the mel spectrogram in the implementor side is a good idea.

The RMVPE repo already has a export script, some other onnx inference implementations already uses that inputs/outputs spec. I think OpenUtau should do the same: https://github.com/yxlllc/RMVPE/blob/main/export.py

I will attach a model with input `waveform` and outputs `f0` and `uv` here: https://mega.nz/file/S5ZXCDQK#S9d_e3Ig7sHCFn1QRB-OiJKjUBudGxlJjCovEtJafvs

This ONNX file packs all mel spectrogram and decoding steps in it.

OpenUtau.Core/Analysis/Rmvpe.cs
```diff
@@ -0,0 +1,342 @@
...
+            ?? throw new InvalidDataException("RMVPE model must expose a uv output.");
+    }
+
+    static float[] ToMono(UWavePart wavePart) {
+        if (wavePart.channels == 1) {
+            return wavePart.Samples;
+        }
+        return Enumerable.Range(0, wavePart.Samples.Length / wavePart.channels)
+            .Select(i => wavePart.Samples.Skip(i * wavePart.channels).Take(wavePart.channels).Average())
+            .ToArray();
```
yqzhishen: Is this an O(n^2) implementation? Avoid using LINQ, use raw arrays, like that in `MidiExtractor`.

OpenUtau.Core/Analysis/Rmvpe.cs
```diff
@@ -0,0 +1,342 @@
...
+        var scale = (double)sourceSampleRate / SampleRate;
+        for (int i = 0; i < targetLength; ++i) {
+            var sourcePos = i * scale;
+            var left = Math.Clamp((int)Math.Floor(sourcePos), 0, samples.Length - 1);
+            var right = Math.Min(left + 1, samples.Length - 1);
+            var frac = sourcePos - left;
+            result[i] = (float)(samples[left] * (1.0 - frac) + samples[right] * frac);
+        }
+        return result;
+    }
```
yqzhishen, line 389: This is not a correct resampling implementation. Both NWaves and NAudio have resampling APIs, please use them.

OpenUtau/Views/MainWindow.axaml.cs
```diff
@@ -1332,25 +1332,39 @@ await MessageBox.Show(this,
...
+                                RmvpeResult? rmvpeResult = null;
+                                if (voicePart != null && transcribeVm.PredictPitd) {
+                                    msgbox.SetText($"{text} {part.name}\nPredicting f0...");
+                                    using var rmvpe = new RmvpeTranscriber();
+                                    rmvpeResult = rmvpe.Infer(wavePart);
+                                }
+                                return (voicePart, rmvpeResult);
                             }
                         });
                     }
```
yqzhishen: Don't mix pitch extraction tasks with MIDI extraction tasks together. Run MIDI extraction first, then run pitch extraction. This reduces memory usage.

OpenUtau.Core/Analysis/Rmvpe.cs
```diff
@@ -0,0 +1,355 @@
...
+        var f0 = f0Tensor.ToArray();
+        var uv = uvTensor.ToArray();
+        if (f0.Length != uv.Length) {
+            throw new InvalidDataException($"Unexpected RMVPE output sizes: f0={f0.Length}, uv={uv.Length}");
+        }
+        for (int i = 0; i < f0.Length; ++i) {
+            if (uv[i]) {
+                f0[i] = 0f;
+            }
+        }
```
yqzhishen: If the f0 curve is treated as a continuous sequence in later processes, a linear interpolation should be done here instead of zero filling. If F0 = 0, log(F0) goes to -inf. Remember to interpolate on the semitone/MIDI scale. So I recommend storing midi pitch instead of f0 array in `RmvpeResult`.

yqzhishen:
I did some local tests, and here are some issues and my suggestions:

1. The checkbox seems to have font size and alignment issues? image
2. This checkbox should be disabled if RMVPE is not installed. Similar to GAME and SOME, a tooltip with guidance for the user to install it through package manager (I will ask stakira to add it to the registry) should be used.

OpenUtau.Core/Analysis/Rmvpe.cs
```diff
@@ -0,0 +1,413 @@
...
+    static string ResolveModelPath() {
+        var candidates = new List<string> {
+            Path.Combine(PathManager.Inst.DependencyPath, "rmvpe", "rmvpe.onnx"),
+            Path.Combine(PathManager.Inst.DependencyPath, "RMVPE", "rmvpe.onnx"),
+            Path.Combine(PathManager.Inst.DependencyPath, "DiffSinger", "rmvpe.onnx"),
+        };
+        var current = new DirectoryInfo(PathManager.Inst.RootPath);
+        for (var i = 0; i < 6 && current != null; ++i, current = current.Parent) {
+            candidates.Add(Path.Combine(current.FullName, "RMVPE", "rmvpe.onnx"));
+        }
```
yqzhishen: Don't scan anything outside OpenUtau directory. Just `Dependencies/rmvpe` is enough.

OpenUtau/Views/TranscribeDialog.axaml.cs
```diff
@@ -20,6 +24,36 @@ void OnCancelClicked(object? sender, RoutedEventArgs e) {
             Close();
         }
 
+        void OnRmvpeRowPressed(object? sender, PointerPressedEventArgs e) {
+            e.Handled = true;
+            _ = HandleRmvpeRowClick();
+        }
```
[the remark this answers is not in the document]
yqzhishen: I think this is not important.

OpenUtau/Views/MainWindow.axaml.cs
```diff
@@ -1316,9 +1316,6 @@ await MessageBox.Show(this,
                 try {
                     string text = ThemeManager.GetString("context.part.transcribing");
                     var msgbox = MessageBox.ShowModal(this, $"{text} {part.name}", text);
-                    var scheduler = TaskScheduler.FromCurrentSynchronizationContext();
-
-                    Task<UVoicePart?> transcribeTask;
                     Func<bool> confirmLongChunk = () => {
                         return Dispatcher.UIThread.InvokeAsync(async () => {
```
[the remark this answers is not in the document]
yqzhishen: Another issue: if the transcribing... dialog is manually closed, the task still runs and generates a voice part.

I think these stuff related to cancellation or interruption should go with another PR, because some more complicated refactoring may be needed.

emeraldsingers:
oh shit, missclicked. but nothing changed

PR #1905, opened 2026-01-04
Add Curve Tools and Curve Selection/Copy/Paste

maiko3tattun:
~~Please merge #1811 first 🙏~~ ✅ 

## New Features:
- Add CurveSelectTool, CurvePenTool, CurveEraserTool
  - Selecting a curve will deselect the note.
  - The selection range affects the snap state of the pianoroll.
  - The Curve Eraser Tool now allows you to reset curves with a left click, improving usability on laptops.
- Add Copy/Cut/Paste Curve function
  - The last item copied from either the notes or the curve will be pasted.

## Code Refactoring:
- Compact code for setting curves within a specified range

## Not Yet:
- ~~I couldn't figure out how to set the background color for the tool selector~~
- The Line Tool is currently not available

PR #2008, opened 2026-03-15
Toolbar Enhancement

maiko3tattun:
Please merge #1905 first 🙏

## Background of Development
- There was a high demand for customizing editing tools, but the proliferation of tools had led to increased complexity. Furthermore, in an effort to expand functionality without adding more tools, we implemented operations using complex modifier keys, which made the tools difficult to use.
- Internally, the toolbar did not simply select an item from a list; instead, each button acted as a toggle, resulting in convoluted code.
- The goal here is to improve usability and future scalability through UI reorganization and refactoring.

## New Features
- In the piano roll toolbar, similar tools have been grouped together, giving it a cleaner look. You can access related tools by long-pressing or right-clicking a button (just like in Adobe).
- The tool's previous setting is now saved. For example, if a user previously used the PenPlusTool instead of the PenTool, the PenPlusTool will be selected automatically—without any special settings—when the second button from the left is pressed or the 2 key is pressed.

https://github.com/user-attachments/assets/c9b37ce0-9ed5-41d6-9d80-0cf6ad8ed6a3

## Refactoring
- EditTool class has been implemented, making it easier to retrieve the currently selected tool.
- To enable dynamic switching of tool icons, the path has been moved to the style file.

## Removed Features
- The shortcut keys previously assigned to individual sub-tools have been unified by group.
- The “Use Pen Plus Tool by default” setting has been removed.


OpenUtau.Core/Util/EditTool.cs
rokujyushi, line 1: ファイル名とクラス名を揃えてください。
Please make sure the file name and the class name match.

stakira (approved):

PR #2009, opened 2026-03-17
CoreML fixes

AnAndroNerd:
Issues can happen do to CoreML on MacOS, this fixes said issues by

- Inferencing the diffsinger accoustic model on the CPU
- Running the GAME encoder model on CPU

### DIFF
--- a/OpenUtau.Core/Analysis/Game.cs
+++ b/OpenUtau.Core/Analysis/Game.cs
@@ -6,6 +6,7 @@ using System.Text.Json;
 using System.Text.Json.Serialization;
 using Microsoft.ML.OnnxRuntime;
 using Microsoft.ML.OnnxRuntime.Tensors;
+using OpenUtau.Core.Util;
 using Serilog;
 
 namespace OpenUtau.Core.Analysis;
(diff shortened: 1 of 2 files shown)


OpenUtau.Core/Analysis/Game.cs
```diff
@@ -96,10 +97,10 @@ public Game(string? location) {
     /// </summary>
     private void EnsureSessionsLoaded() {
         if (sessionsLoaded) return;
-        encoderSession = CreateSession("encoder.onnx");
```
stakira, line 99: have you thought about user on other platforms?

AnAndroNerd: The issue here boiled down to how CoreML bounces things back and forth between the neural engine and the CPU on Mac, DirectML doesn't do this and neither does NNAPI so it checks if we're using CoreML and if we are, it runs that task on the CPU. You can only select CoreML on Mac, this felt like the cleanst way to check.

stakira: I mean, when you write `Preferences.Default.OnnxRunner == "CoreML"`, you are leaving others out.

stakira: The logic is actually right. CoreML use cpu and others use preference.

PR #2011, opened 2026-03-18
[KR C+V] Request to add Phonemizer for new voice bank format Korean C+V 

SODAsoo07:
It is a new format C+V ponemizer that separates and synthesizes Korean Consonant/Vowel/Coda.


SODAsoo07:
I've added a basic pronunciation converter function to the code. Now, when you type "words that are orthogonal, but their actual pronunciation doesn't match" into the note, the phonemes are parsed to the standard Korean pronunciation. (An exceptional case where automatic parsing and actual correct pronunciation do not match can be customized by writing down the word-pronunciation you want through the kr-CpV-dict.yaml file.)

PR #2014, opened 2026-03-19
SBP Update and Fixes

Cadlaxa:
- Implemented pre-utterance alias length (reads the alias oto parameters `from left blank to preutter`) every child SBP phonemizer will implement this to their end:
```
// true: endings end in a 50-tick gap
// false: uses the alias length instead
protected override bool NoGap => true;

protected override double GetTransitionBasicLengthMs(string alias, int tone, PhonemeAttributes attr) {
    double otoLength = GetTransitionBasicLengthMsByOto(alias, tone, attr);
    return otoLength;
}
```
- Fixed pitch suffix bug in AssignAllAffixes where it assigns wrong pitch to tones other than vowelTone
- Implemented IsGlide to snap certain alias to the note's starting position
- Implemented Global Dictionary
- Implemented diphthong splits via yaml
- Implemented vowel sustains via yaml
- Small code refactor


OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -1047,20 +1024,30 @@ private string CheckVCExceptions(string vc) {
...
                 alias = alias.Replace(consonant, "3");
             }
 
             return alias;
         }
+
+        // Endings has 50 ticks gap
+        protected override bool NoGap => true;
+
+        protected override double GetTransitionBasicLengthMs(string alias, int tone, PhonemeAttributes attr) {
```
AnAndroNerd, line 1109: This code does not work properly with the auto convel, this is because this code runs before any of the math is done.

AnAndroNerd (changes requested):

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
Cadlaxa: The ending cap or the pre-utterance phoneme length?

AnAndroNerd: pre-utterance phoneme length

Cadlaxa: Lemme remove that for vccv

AnAndroNerd: we want it for VCCV, it just doesn't work right currently. look at discord for more info

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -1045,8 +1033,35 @@ protected override string ValidateAlias(string alias, int tone = 0) {
...
+            else
+                finalConvel = baseConvel + (100 - (100 * ((float)un.duration / 480)));
+
+            return Math.Clamp(finalConvel, velMin, velMax);
+        }
+
         protected override double GetTransitionBasicLengthMs(string alias, int tone, PhonemeAttributes attr) {
             double otoLength = GetTransitionBasicLengthMsByOto(alias, tone, attr);
+
+            if (useConvel && unotes.Count > 0) {
```
AnAndroNerd: shouldn't use unotes and should use uphonemes somehow

AnAndroNerd (changes requested):

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
AnAndroNerd: this goes for everything in that group

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -288,90 +266,84 @@ string Alt(IEnumerable<string> symbols) =>
...
+                string sym = phonemeSymbols[i];
+                float autoVel;
+                switch (Classify(sym)) {
                     case "V C": case "VC": case "VC-":
                     case "VCC": case "VCC-": case "codaCC": case "C C":
-                    case "VC C":
-                        if (GetNoteForPhoneme(phoneme, notes).lyric == "+" ||
-                            GetNoteForPhoneme(phoneme, notes).lyric == "+~") {
-                            vel = noteVel;
+                    case "VC C": case "V-": case "V CC":
```
AnAndroNerd: should be `case "VC C": case "V-": case "V CC": case "CC-": case "CC":`

AnAndroNerd (changes requested):

AnAndroNerd (approved):

AnAndroNerd (changes requested):
VCCV still isn't right, I will fix it myself

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -288,90 +266,84 @@ string Alt(IEnumerable<string> symbols) =>
...
-                var un = unotes.LastOrDefault(n => n.position <= absPos) ?? unotes[0];
-                int idx = unotes.IndexOf(un);
-                return (un, idx + 1 < unotes.Count ? unotes[idx + 1] : null);
-            }
-            
-            var (curUN, nextUN) = UNoteAt(notes[0].position);
-            int curIdx = unotes.IndexOf(curUN);
-            var prevUN = curIdx > 0 ? unotes[curIdx - 1] : null;
+            UNote curUN = null;
+            if (unotes.Count > 0) {
```
AnAndroNerd: all of these shouldn't use the phoneme notes, but instead unotes so that + notes work right. this also fixes other issues down the line

Image

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -288,90 +266,84 @@ string Alt(IEnumerable<string> symbols) =>
...
+                string sym = phonemeSymbols[i];
+                float autoVel;
+                switch (Classify(sym)) {
                     case "V C": case "VC": case "VC-":
                     case "VCC": case "VCC-": case "codaCC": case "C C":
-                    case "VC C":
-                        if (GetNoteForPhoneme(phoneme, notes).lyric == "+" ||
-                            GetNoteForPhoneme(phoneme, notes).lyric == "+~") {
-                            vel = noteVel;
+                    case "VC C": case "V-": case "V CC": case "CC-": case "CC":
```
AnAndroNerd: `case "CC"` never happens (sorry!)

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -288,90 +266,84 @@ string Alt(IEnumerable<string> symbols) =>
...
+                        if (notes[0].lyric == "+" || notes[0].lyric == "+~") {
+                            autoVel = noteVel;
                             break;
                         }
-                        vel = prevVel ?? noteVel;
+                        autoVel = prevVel ?? noteVel;
                         break;
                     case "onsetCC": case "-CC":
-                        vel = nextVel ?? noteVel;
+                        autoVel = prevVel ?? noteVel;
```
AnAndroNerd: this needs a check for isGlide, it should be noteVel if it's enabled and nextVel if it isn't

AnAndroNerd (changes requested):

AnAndroNerd (approved):

PR #2031, opened 2026-03-28
Support Flag Text Input

maiko3tattun:
## New Feature
It supports text input and display for flags.
Surprisingly, I’ve heard that the lack of this feature is the reason users remain with the original UTAU.

You may realize through this feature that unnecessary flags have been outputted up until now.
You can disable this output in the Expression settings.

https://github.com/user-attachments/assets/ccff97ac-d407-42cb-b035-38bee590439d


maiko3tattun:
Ah, of course, this text box is only displayed when using the Classic Renderer.

maiko3tattun:
- Support for option-type flags has been added
- ~~Displays a warning if the flag cannot be parsed~~
image

maiko3tattun:
The default value for `SkipOutputIfDefault` has been set to `true`. Unnecessary flags will be removed from the values passed to the resampler.
- This does not affect existing projects. This value will be set to `true` when creating new projects and when adding new expressions.
- B flags, whose default values vary by resampler, will continue to be output.

PR #2061, opened 2026-04-18
Detached Pianoroll Window Position Fix

Cadlaxa:
- Updated the `WindowClosing` logic in `PianoRollDetachedWindow` to handle the minimized state safely cuz If the user closes the main app while the piano roll is minimized, the original code would save (int)WindowState as Minimized. The next time they launched the app, the piano roll would open, but it would launch completely minimized and out of sight, confusing the user. (the bug is randomly occurs but I caught it this time)


maiko3tattun:
To apply this to both windows, you should write the code in `OpenUtau.Core/Util/WindowSize.cs`. And I believe it has already been written there #1859

PR #2075, opened 2026-04-26
Separate Phonemizers according to the singer type

Cadlaxa:
- Sepates them with `engine: "value"` PhonemizerAttribute
- Fallbacks to `Utau` if there's no attribute added in the phonemizer


Cadlaxa:
image
----
image

maiko3tattun:
Is it necessary to display phonemizers of unsupported types?

Cadlaxa:
> Is it necessary to display phonemizers of unsupported types?

I've thought about this before but old external phonemizers will have issues with this if we hide them so they have to update their phonemizer class attribute to align them to the respective group, so instead of hiding them, I just group them via the singer type (also a pop-up to warn the users when they're using the wrong phonemizer type)

maiko3tattun:
Then, how about moving the supported groups out of the “More” section and up to the next level? Under the current specifications, laptop users are at a disadvantage because they require precise mouse movements.

Cadlaxa:
Interesting idea, I'll account this too

Cadlaxa:
image
image

PR #2084, opened 2026-05-01
Add Phonetic Hints Plugin

ochakochatblanc:
Some WIP code for adding phonetic hints to lyrics as a batch plugin. Fallback behaviour still buggy.


ochakochatblanc:
Fixed the fallback behaviour. Should be ready to merge upon testing @celeb17-kurisutina

PR #2091, opened 2026-05-02
Rearrangeable Tracks

atouu:
Allow rearranging track headers by dragging

https://github.com/user-attachments/assets/a5420971-71e5-4ef8-91a3-ad5dca703229

### DIFF
--- a/OpenUtau.Core/Commands/TrackCommands.cs
+++ b/OpenUtau.Core/Commands/TrackCommands.cs
@@ -87,6 +87,30 @@ namespace OpenUtau.Core {
         }
     }
 
+    public class SetTrackNoCommand : TrackCommand {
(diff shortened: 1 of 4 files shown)


stakira:
This is great! Mind change the display to show between which tracks it will drop instead? Like this?
image
For simplicity you can render that as part of the hovered track header

atouu:
That's actually the first behaviour but I can't get the pointer to align with the line. I'll check again later

stakira:
> That's actually the first behaviour but I can't get the pointer to align with the line. I'll check again later

Simplest way it's just check in the hovered track header, see if it is in its upper half or lower half, then just draw a line on its top or bottom.

atouu:
I did it @stakira

PR #2092, opened 2026-05-03
Add ENG flag switching to Voice Color remapping

Electric-wind:
This change adds support for changing the ENG flag from the Voice Color remapping dialog.

It allows users to switch to a more suitable resampler for each voice color while remapping voice colors at the same time. This makes resampler selection and testing much faster and more convenient.

Because the list is automatically populated from the resamplers in the Resamplers folder, users no longer need to type resampler names into the ENG flag repeatedly.

I believe this change will make UTAU singer tuning more convenient and efficient.

### DIFF
--- a/OpenUtau/ViewModels/VoiceColorMappingViewModel.cs
+++ b/OpenUtau/ViewModels/VoiceColorMappingViewModel.cs
@@ -1,4 +1,5 @@
-﻿using System;
+using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using DynamicData;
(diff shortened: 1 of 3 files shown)


Electric-wind:
> This change adds support for changing the ENG flag from the Voice Color remapping dialog.
> 
> It allows users to switch to a more suitable resampler for each voice color while remapping voice colors at the same time. This makes resampler selection and testing much faster and more convenient.
> 
> Because the list is automatically populated from the resamplers in the Resamplers folder, users no longer need to type resampler names into the ENG flag repeatedly.
> 
> I believe this change will make UTAU singer tuning more convenient and efficient.

image

maiko3tattun:
Isn't it confusing to have multiple roles mixed together in a single dialog? 🤔

PR #2094, opened 2026-05-03
Fix mono audio bug

ochakochatblanc:
**Mixdown Mono Audio Bug**
- Current implementation results in the 'Mixdown To Wave' audio render to produce a Mono output, in which only the left channel contributes to the Mono output
-  This results in audio panning information entered by the user from being lost, and for imported audio tracks which utilise both channels from being improperly mixed down when rendering.
- Render Waves per Track also faces issue of being improperly mixed down (Left channel contribution)

Solution:
- 'Mixdown To Wave' results in stereo output as to align with how OpenUTAU renders the audio within the track player.
- 'Render To Files' is rendered as Mono, but both Left and Right channels contribute to the output for any edge cases where both channels are necessary for the correct output.

### DIFF
--- a/OpenUtau.Core/PlaybackManager.cs
+++ b/OpenUtau.Core/PlaybackManager.cs
@@ -320,7 +320,7 @@ namespace OpenUtau.Core {
                     DocManager.Inst.ExecuteCmd(new ProgressBarNotification(0, $"Exporting to {exportPath}."));
 
                     CheckFileWritable(exportPath);
-                    WaveFileWriter.CreateWaveFile16(exportPath, new ExportAdapter(projectMix).ToMono(1, 0));
+                    WaveFileWriter.CreateWaveFile16(exportPath, new ExportAdapter(projectMix));
                     DocManager.Inst.ExecuteCmd(new ProgressBarNotification(0, $"Exported to {exportPath}."));
                 } catch (IOException ioe) {
                     var customEx = new MessageCustomizableException($"Failed to export {exportPath}.", $"<translate:errors.failed.export>: {exportPath}", ioe);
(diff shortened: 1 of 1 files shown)


celeb17-kurisutina (commented):
## Code Review

### Change 1 (RenderMixdown): ✅ Correct

Removing `.ToMono(1, 0)` is the right fix. The mixdown applies per-track fader + panning, so L and R channels differ. `.ToMono(1, 0)` drops the entire right channel, losing all panning info. Exporting full stereo preserves it.

### Change 2 (RenderToFiles): ❌ Not needed — introduces clipping

`RenderTracks` only processes **UVoicePart** (UWavePart is excluded by `PrepareRequests()`). Voice parts are mono (`channels=1`). `WaveSource.Mix` writes each mono sample to both L and R positions equally, so L == R.

`.ToMono(1, 0)` takes the left channel, which is identical to the right — no information is lost.

Changing to `.ToMono(1, 1)` **sums** L + R at full volume. For a centered voice at 0.7 amplitude on both channels, the mono sum becomes 0.7 + 0.7 = 1.4 → **hard clipping**. This produces audible distortion on any content where both channels are active.

**Recommendation:** Revert this change. The original `.ToMono(1, 0)` is correct for mono voice parts.

OpenUtau.Core/PlaybackManager.cs
```diff
@@ -349,7 +349,7 @@ await Task.Run(() => {
                         DocManager.Inst.ExecuteCmd(new ProgressBarNotification(0, $"Exporting to {file}."));
 
                         CheckFileWritable(file);
-                        WaveFileWriter.CreateWaveFile16(file, new ExportAdapter(trackMixes[i]).ToMono(1, 0));
+                        WaveFileWriter.CreateWaveFile16(file, new ExportAdapter(trackMixes[i]).ToMono(1, 1));
```
celeb17-kurisutina, line 352: This change introduces clipping. `RenderTracks` only processes UVoicePart, which are mono (`channels=1`). `WaveSource.Mix` writes each mono sample to both L and R positions equally (L == R), so `.ToMono(1, 0)` takes the left channel which is identical to the right — no information is lost.

Changing to `.ToMono(1, 1)` **sums** L + R at full volume. For a centered voice at 0.7 amplitude on both channels, the mono sum becomes 0.7 + 0.7 = 1.4 → **hard clipping**.

Revert this to `.ToMono(1, 0)` — it's correct for mono voice parts.

stakira:
Since there is no reponse, I pushed the correct fix.

ochakochatblanc:
ty

PR #2096, opened 2026-05-03
Add Keyboard Shortcut Remapper

Cadlaxa:
- Remaps exisitng keyboard shortcuts in preferences
- Pianoroll toggles also shows the remapped key shortcut (no more hardcoded strings)


Cadlaxa:
image

- left click then click any key to remap them
- right click to reset individual keys

maiko3tattun:
- In macOS, the Command and Control keys have been swapped. For example, the search shortcut used to be Command+F, but in this PR, it has been changed to Control+F.
- In PianorollViewModel, “Hotkeys” must be a Dictionary<string, KeyGesture>. The text displayed in the menu simply needs to be parsed into KeyGesture.

maiko3tattun:
It appears that the current version for macOS has a bug where the Ctrl and Command keys are labeled in the menu in the opposite order of their actual behavior. It seems that fixing the two issues I mentioned earlier will naturally resolve this bug as well 👍

Cadlaxa:
Thanks for reporting since I don't have any macos device to test this properly 😔

maiko3tattun:
I found a few other issues as well 🥲
This might require a major overhaul. I’ll post what I’ve been working on here for now.
https://github.com/Cadlaxa/OpenUtau/pull/6

Cadlaxa:
This PR is ready to merged, Maiko already checked and fixed the issues for Macos and other key bindings

Cadlaxa:
Added shortcuts for the new line tools

maiko3tattun:
Come to think of it, we should probably discuss whether or not to assign default key bindings to all the batch editing functions. Since not all of these functions are frequently used, perhaps only users who wish to use them should customize the keys. Instead, we might add other shortcut in the future—such as transport controls—that are likely to be used more frequently.

PR #2111, opened 2026-05-09
Native Multi-Phonemizer per Track

Cadlaxa:
- Support for overriding phonemizer from the selected notes
- Toggle to hide the phonemizer tag

### DIFF
--- a/OpenUtau/ViewModels/NotePropertiesViewModel.cs
+++ b/OpenUtau/ViewModels/NotePropertiesViewModel.cs
@@ -45,9 +45,7 @@ namespace OpenUtau.App.ViewModels {
                 if (isDefault) {
                     if (Part == null) return "Default";
                     var track = DocManager.Inst.Project.tracks[Part.trackNo];
-                    string trackId = !string.IsNullOrEmpty(track.phonemizer) 
-                        ? track.phonemizer 
-                        : track.Phonemizer.GetType().FullName ?? "";
+                    string trackId = track.Phonemizer.GetType().FullName ?? "";
                     return $"Default ({GetPhonemizerDisplayName(trackId)})";
                 }
                 var factory = OpenUtau.Api.PhonemizerFactory.GetAll().FirstOrDefault(f => 
@@ -265,9 +263,7 @@ namespace OpenUtau.App.ViewModels {
             string defaultHeader = "Default";
             if (Part != null) {
                 var track = DocManager.Inst.Project.tracks[Part.trackNo];
-                string trackId = !string.IsNullOrEmpty(track.phonemizer) 
-                    ? track.phonemizer 
-                    : (track.Phonemizer?.GetType().FullName ?? string.Empty);
+                string trackId = track.Phonemizer.GetType().FullName ?? "";
                 defaultHeader = $"Default ({GetPhonemizerDisplayName(trackId)})";
             }
             items.Add(new MenuItemViewModel() {
@@ -411,13 +407,13 @@ namespace OpenUtau.App.ViewModels {
                         VibratoVolLink = note.vibrato.volLink;
                         this.RaisePropertyChanged(nameof(VibratoVolLink));
                     }
-                } else if (cmd is ChangeNotePhonemizerCommand) {
+                }/* else if (cmd is ChangeNotePhonemizerCommand) {
                     PhonemizerOverride = note.PhonemizerOverride ?? "";
                     this.RaisePropertyChanged(nameof(PhonemizerOverride));
                     if (Part != null) {
                         DocManager.Inst.Project.Validate(new ValidateOptions { Part = Part });
                     }
-                }
+                }*/
             } else if (cmd is ExpCommand) {
                 if (cmd is PitchExpCommand) {
                     if (note.pitch.data.Count >= 2) {
@@ -435,9 +431,6 @@ namespace OpenUtau.App.ViewModels {
             } else if (cmd is NotePresetChangedNotification) {
                 PortamentoPresets = new ObservableCollection<NotePresets.PortamentoPreset>(NotePresets.Default.PortamentoPresets);
                 VibratoPresets = new ObservableCollection<NotePresets.VibratoPreset>(NotePresets.Default.VibratoPresets);
-            } else if (cmd is TrackChangePhonemizerCommand) {
-                RefreshPhonemizers();
-                this.RaisePropertyChanged(nameof(PhonemizerOverrideText));
             }
         }
         #endregion
--- a/OpenUtau/ViewModels/NotePropertiesViewModel.cs
+++ b/OpenUtau/ViewModels/NotePropertiesViewModel.cs
@@ -11,6 +11,7 @@ using OpenUtau.Core.Util;
 using ReactiveUI;
 using ReactiveUI.Fody.Helpers;
 using SharpCompress;
+using OpenUtau.Api;
 
 namespace OpenUtau.App.ViewModels {
     public class NotePropertiesViewModel : ViewModelBase, ICmdSubscriber {
@@ -34,7 +35,18 @@ namespace OpenUtau.App.ViewModels {
         [Reactive] public float AutoVibratoNoteLength { get; set; }
         [Reactive] public bool AutoVibratoToggle { get; set; }
         [Reactive] public bool IsNoteSelected { get; set; } = false;
+        [Reactive] public IReadOnlyList<MenuItemViewModel>? PhonemizerMenuItems { get; set; }
+        public ReactiveCommand<string?, System.Reactive.Unit> SelectPhonemizerCommand { get; }
+        public string PhonemizerOverrideText => string.IsNullOrEmpty(PhonemizerOverride) ? "Default" : PhonemizerOverride;
 
+        private string? _phonemizerOverride;
+        public string? PhonemizerOverride {
+            get => _phonemizerOverride;
+            set {
+                this.RaiseAndSetIfChanged(ref _phonemizerOverride, value);
+                this.RaisePropertyChanged(nameof(PhonemizerOverrideText));
+            }
+        }
         [Reactive] public ObservableCollection<NotePresets.PortamentoPreset>? PortamentoPresets { get; private set; }
         public NotePresets.PortamentoPreset? ApplyPortamentoPreset {
             get => appliedPortamentoPreset;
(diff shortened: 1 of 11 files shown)


Cadlaxa:
accessible via notes panel to change the selected note's phonemizer
image

yqzhishen:
Consecutive phonemizer changes don't show language name on each notes. Is it by design?
image

Cadlaxa:
> Consecutive phonemizer changes don't show language name on each notes. Is it by design?
> image
> 

Oh yes, the phonemizer tag only shows in the overridden language transition, doing it for every notes feels cluttered

yqzhishen:
> > Consecutive phonemizer changes don't show language name on each notes. Is it by design?
> > image
> > 
> 
> Oh yes, the phonemizer tag only shows in the overridden language transition, doing it for every notes feels cluttered

If so, should a language tag be added on the note which changes the phonemizer back to default?

Cadlaxa:
Good point since it's a language transition too, lemme edit some changes

yqzhishen:
A tiny detail here. Slur notes and syllable continuation notes should follow its main note and disable the phonemizer dropdown.

image

yqzhishen:
Seems like the language tags aren't updated if the default phonemizer is changed from the track window

Cadlaxa:
Yeah, the tags got delayed sometimes, I'll fix this tomorrow too and also the bug with the current phonemizer transition tag 😔

OpenUtau/ViewModels/NotePropertiesViewModel.cs
```diff
@@ -34,7 +35,42 @@ public class NotePropertiesViewModel : ViewModelBase, ICmdSubscriber {
...
+                if (isDefault && Part != null && DocManager.Inst?.Project?.tracks != null && Part.trackNo < DocManager.Inst.Project.tracks.Count) {
+                    var track = DocManager.Inst.Project.tracks[Part.trackNo];
+                    targetId = track.phonemizer ?? "";
+                    if (string.IsNullOrEmpty(targetId) && track.Phonemizer != null) {
+                        targetId = track.Phonemizer.GetType().FullName ?? "";
+                    }
+                }
+                if (string.IsNullOrEmpty(targetId)) return "Default";
+                var factory = OpenUtau.Api.PhonemizerFactory.GetAll().FirstOrDefault(f => 
+                    f.name == targetId || f.GetType().FullName == targetId || f.GetType().Name == targetId);
```
yqzhishen: `f.GetType()` returns the type of the factory object itself, not the phonemizer type. Please use `f.type`, so that the phonemizer name can display correctly (instead of triggering a regex fallback that inserts spaces before every uppercase letter, making "DiffSinger" to become "Diff Singer").

Same thing happens on Line 268.

Image

OpenUtau/ViewModels/NotePropertiesViewModel.cs
```diff
@@ -318,6 +433,12 @@ public void OnNext(UCommand cmd, bool isUndo) {
             } else if (cmd is NotePresetChangedNotification) {
                 PortamentoPresets = new ObservableCollection<NotePresets.PortamentoPreset>(NotePresets.Default.PortamentoPresets);
                 VibratoPresets = new ObservableCollection<NotePresets.VibratoPreset>(NotePresets.Default.VibratoPresets);
+            } else if (cmd is ChangeNotePhonemizerCommand) {
+                PhonemizerOverride = note.PhonemizerOverride ?? "";
+                this.RaisePropertyChanged(nameof(PhonemizerOverride));
+                if (Part != null) {
+                    DocManager.Inst.Project.Validate(new ValidateOptions { Part = Part });
+                }
             }
```
yqzhishen: 1. `ChangeNotePhonemizerCommand` extends `NoteCommand`, so this should move into a `else if` branch above in the `if (cmd is NoteCommand)` block. Otherwise it's dead code, so the dropdown isn't updated correctly.
2. You need to handle `TrackChangePhonemizerCommand` here so the dropdown can update on default phonemizer change from track:
```c#
else if (cmd is TrackChangePhonemizerCommand) {
    RefreshPhonemizers();
    this.RaisePropertyChanged(nameof(PhonemizerOverrideText));
}
```

After these fixes, the dropdown gets updated properly.

OpenUtau/ViewModels/NotePropertiesViewModel.cs
yqzhishen: Also this seems to be the root cause why the language name tag of default phonemizer isn't refreshed after track phonemizer change before you use a dot character.

yqzhishen:
~~There is still a bug causing phonemizer name on note property panel to stuck without updating. Debugged for hours without any clue, maybe I need more time to investigate.~~

I find the root cause now. I will let DeepSeek to comment on behalf of me.

OpenUtau/ViewModels/NotePropertiesViewModel.cs
```diff
@@ -34,7 +35,39 @@ public class NotePropertiesViewModel : ViewModelBase, ICmdSubscriber {
...
+        public ReactiveCommand<string?, System.Reactive.Unit> SelectPhonemizerCommand { get; }
+        [Reactive] public bool IsPhonemizerEnabled { get; set; } = true;
+        public string PhonemizerOverrideText {
+            get {
+                string? targetId = PhonemizerOverride;
+                bool isDefault = string.IsNullOrEmpty(targetId);
+                if (isDefault) {
+                    if (Part == null) return "Default";
+                    var track = DocManager.Inst.Project.tracks[Part.trackNo];
+                    string trackId = !string.IsNullOrEmpty(track.phonemizer)
```
yqzhishen: `track.phonemizer` (lowercase, string) is the YAML serialization field — `BeforeSave()` writes it and `AfterLoad()` reads it, then clears it. `track.Phonemizer` (uppercase, `[YamlIgnore]`) is the runtime object. The getter checks the serialization string first via `!string.IsNullOrEmpty(track.phonemizer)`, but this string can be stale: autosave triggers `BeforeSave()`, which writes the current phonemizer type into `track.phonemizer`. Later, when `TrackChangePhonemizerCommand` changes `track.Phonemizer`, the string field is not updated, so the getter still returns the old phonemizer name. This is the root cause of the "delayed update" bug.

**Fix:** just read `track.Phonemizer.GetType().FullName` directly and skip the `track.phonemizer` check entirely. The runtime object is always initialized.

OpenUtau/ViewModels/NotePropertiesViewModel.cs
```diff
@@ -196,11 +243,75 @@ public void LoadPart(UPart? part) {
...
+            return !string.IsNullOrEmpty(displayName) ? displayName : factory.ToString();
+        }
+
+        public void RefreshPhonemizers() {
+            var items = new List<MenuItemViewModel>();
+
+            string defaultHeader = "Default";
+            if (Part != null) {
+                var track = DocManager.Inst.Project.tracks[Part.trackNo];
+                string trackId = !string.IsNullOrEmpty(track.phonemizer)
```
yqzhishen: Same issue as above — reads `track.phonemizer` (serialization string) first, which can be stale after autosave. Should read `track.Phonemizer.GetType().FullName` directly.

OpenUtau/ViewModels/NotePropertiesViewModel.cs
```diff
@@ -318,6 +435,9 @@ public void OnNext(UCommand cmd, bool isUndo) {
             } else if (cmd is NotePresetChangedNotification) {
                 PortamentoPresets = new ObservableCollection<NotePresets.PortamentoPreset>(NotePresets.Default.PortamentoPresets);
                 VibratoPresets = new ObservableCollection<NotePresets.VibratoPreset>(NotePresets.Default.VibratoPresets);
+            } else if (cmd is TrackChangePhonemizerCommand) {
```
yqzhishen: This handler is **dead code** when no notes are selected. `var note = selectedNotes.FirstOrDefault()` at line 360 returns null → `if (note == null) return;` → the method returns before reaching this handler. The result: when the user switches the track phonemizer without selecting any notes, the dropdown menu never rebuilds and the button text never updates.

**Fix:** move this handler above line 360, before the `note == null` check. Also add `return;` after it so it doesn't fall through to the note-dependent code below.

OpenUtau/ViewModels/NotePropertiesViewModel.cs
```diff
@@ -300,6 +411,12 @@ public void OnNext(UCommand cmd, bool isUndo) {
                         VibratoVolLink = note.vibrato.volLink;
                         this.RaisePropertyChanged(nameof(VibratoVolLink));
                     }
+                } else if (cmd is ChangeNotePhonemizerCommand) {
+                    PhonemizerOverride = note.PhonemizerOverride ?? "";
+                    this.RaisePropertyChanged(nameof(PhonemizerOverride));
```
yqzhishen: Redundant — the `PhonemizerOverride` setter (line 67) already fires `PropertyChanged("PhonemizerOverride")` via `RaiseAndSetIfChanged`. Additionally, `PhonemizerOverride` is not bound to any XAML element (only `PhonemizerOverrideText` is bound, on the Button's `Content`), so this event has no listeners.

yqzhishen: Don't comment out the whole block. Just the line `this.RaisePropertyChanged(nameof(PhonemizerOverride));` is redundant, though it doesn't really matter

OpenUtau/ViewModels/NotePropertiesViewModel.cs
```diff
@@ -435,9 +431,6 @@ public void OnNext(UCommand cmd, bool isUndo) {
             } else if (cmd is NotePresetChangedNotification) {
                 PortamentoPresets = new ObservableCollection<NotePresets.PortamentoPreset>(NotePresets.Default.PortamentoPresets);
                 VibratoPresets = new ObservableCollection<NotePresets.VibratoPreset>(NotePresets.Default.VibratoPresets);
-            } else if (cmd is TrackChangePhonemizerCommand) {
-                RefreshPhonemizers();
-                this.RaisePropertyChanged(nameof(PhonemizerOverrideText));
```
yqzhishen: Don't delete this, it needs to move to the very beginning of this method, before `if (note == null) { return; }`. Otherwise, the note property panel can't update if no notes are selected when the track phonemizer changes.

stakira:
Is this mainly for diffsinger?

Cadlaxa:
It works fine for every types of banks since it just pulls the existing phonemizer, transitions are good too even in concat ****

PR #2112, opened 2026-05-09
[VCCV EN] Automattic Convel

AnAndroNerd:
See "timings" [here](https://github.com/ExpressiveLabs/utau-handbook/blob/main/docs/vccv%20doc.md) for info on convel and vccv.

### DIFF
--- a/OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
+++ b/OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
@@ -217,19 +217,30 @@ namespace OpenUtau.Plugin.Builtin {
             }
             return phoneme;
         }
+        
+        
+        private List<UNote> unotes = new();
 
+        public override void SetUp(Note[][] notes, UProject project, UTrack track) {
+            base.SetUp(notes, project, track);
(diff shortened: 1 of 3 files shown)


OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -216,6 +225,154 @@ private string ReplacePhoneme(string phoneme, int tone) {
...
+                (new Regex($@"^{V}{C} {C}$"), "VC C"),
+                (new Regex($@"^{V}{C}$"),     "VC"),
+                (new Regex($@"^{C}{C2}$"),    "onsetCC"),
+                (new Regex($@"^{C}{C}$"),     "codaCC"),
+                (new Regex($@"^{C}{V}$"),     "CV"),
+                (new Regex($@"^{V}$"),        "V"),
+            };
+
+            string Classify(string alias) {
+                if (starlightccs.Contains(alias)) return "codaCC";
```
AnAndroNerd, line 294: add a comment saying something like "[ll], [mm] and [nn] are not coda ccs, but the logic is still correct"

PR #2119, opened 2026-05-12
Render Live Waveforms

Cadlaxa:
- Uses the already rendered cache to generate the waveforms on the fly instead waiting the final renders in the memory


OpenUtau/Controls/WaveformImage.cs
```diff
@@ -78,32 +81,96 @@ public override void Render(DrawingContext context) {
...
+                        
                         if (sampleData.Length < sampleCount) {
                             Array.Resize(ref sampleData, sampleCount);
                         }
+                        
+                        bool needsAnotherFrame = false;
                         Array.Clear(sampleData, 0, sampleData.Length);
-                        part.Mix.Mix(samplePos, sampleData, 0, sampleCount);
+                        
+                        /*if (OpenUtau.Core.PlaybackManager.Inst.StartingToPlay ||
```
stakira: Make sure you finish the PR and cleanup.

nagotown:
Would it be possible to have an option to disable the animation on update? Or maybe something more subtle like SynthV's?

PR #2126, opened 2026-05-15
New phonemizer: Norwegian to Japanese (NO to JA)

LowAhBeepOh:
This adds a phonemizer that converts Norwegian (Bokmål) words/lyrics into aliases/phonemes that Japanese voicebanks can understand, making it possible for those voicebanks to sing Norwegian.


LowAhBeepOh:
closing this since norwegian isn't really used in utau, and because nothing has happened in this PR for around 4 months now

PR #2135, opened 2026-05-20
Clamp variance delta outputs for diffsinger acoustics&fix build.

KakaruHayate:
Clamp computed variance delta arrays for acoustic features before creating ONNX input tensors. Energy, breathiness, and voicing are clamped to [-96f, 0f], and tension is clamped to [-10f, 10f]. This prevents extreme/out-of-range values from being passed to the model and helps stabilize rendering in DiffSingerRenderer.cs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,10 +49,8 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 20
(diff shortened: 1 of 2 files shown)


KakaruHayate:
@stakira

celeb17-kurisutina (approved):

PR #2139, opened 2026-05-21
Envelope Inversion and Overlap Crushing for Negative Preutter

Cadlaxa:
- when an alias has a negative preutterance, the envelope calculation fails to account for the shifted timeline. This causes two distinct visual and audible bugs:
    1. The Crushing Effect: The leading alias calculates its tailIntrude assuming the delayed alias starts at 0ms
    2. Self-Inversion: Short notes with heavily negative preutterances mathematically try to start after their own physical duration has ended, causing the note to invert its own envelope.

Fix: dynamically respect and push back against delayed attack points.


Cadlaxa:
negative preutter

PR #2142, opened 2026-05-21
Ripple Notes When Resizing using Shift Key

hilmiyafia:
Here's the demo: https://discord.com/channels/551606189386104834/876772238681071656/1345358962131992577

This feature is also on the original Utau

### DIFF
--- a/OpenUtau/Controls/PianoRoll.axaml.cs
+++ b/OpenUtau/Controls/PianoRoll.axaml.cs
@@ -690,6 +690,7 @@ namespace OpenUtau.App.Controls {
                     editState = new NoteResizeEditState(
                         control, ViewModel, this, noteHitInfo.note,
                         args.KeyModifiers == KeyModifiers.Alt,
+                        args.KeyModifiers == KeyModifiers.Shift,
                         fromStart: noteHitInfo.hitResizeAreaFromStart);
                     Cursor = ViewConstants.cursorSizeWE;
                 } else if (args.KeyModifiers == cmdKey) {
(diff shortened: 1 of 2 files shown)


rokujyushi (changes requested):
Regarding key input, please use member variables instead of arguments.
[NoteEditState](https://github.com/openutau/OpenUtau/pull/2179/changes#diff-5e873b045729574caad41ef00394752b22d483a12e60f5f69124f6e0a2ce12c7L44) has member variables for accepting key input and other flags.

hilmiyafia:
> Regarding key input, please use member variables instead of arguments. [NoteEditState](https://github.com/openutau/OpenUtau/pull/2179/changes#diff-5e873b045729574caad41ef00394752b22d483a12e60f5f69124f6e0a2ce12c7L44) has member variables for accepting key input and other flags.

@rokujyushi 

The member variables are not immediately updated for the first resize note update method call, this is causing a gap.

Do you know how to make the shiftHeld member variable more real-time?

https://github.com/user-attachments/assets/ed2e43b3-8e3a-400a-9db1-387118162e9c

rokujyushi:
You do not need to close the pull request every time there are minor changes. If you wish to withdraw the review, please change it to a draft.

rokujyushi:
Do you delete your branch every time you fix the source code?
While you might recreate a branch to clean up the commit history, please do not delete branches that are still being fixed or in progress.
Doing so makes it impossible to review the modified source code.

rokujyushi:
image
This is a rough change, but with this approach, the key input will always be checked whenever an Update occurs.
Please refactor this based on the reference.

hilmiyafia:
@rokujyushi I'm sorry, I won't do it again next time. Thank you for the suggestion.

I already use shiftHeld in the update method. I didn't only read its value on the constructor.

What you see in the wrong behavior part in the video is exactly caused by the code like in your screenshot, by using the shiftHeld in the update method.

Here is the new branch https://github.com/hilmiyafia/openutau/tree/ripple

PR #2144, opened 2026-05-21
Changes and improvements on rule-based Filipino phonemizer

UtaUtaUtau:
This PR includes a few improvements on the G2P to follow the behavior of the language more closely.

Main changes:
 - Added glottal stop insertion rule between vowels
 - Added vowel to glide conversion for certain vowel combinations (u + vowel, i + vowel)
 - Added sh digraph
 - Added coda-exclusive phonemes for diphthongs (Y, W)
Extra:
 - `|` is now considered a bypass character for these extra rules. For example, tatsulok is phonemized as [t a ty u l o k] which is wrong, but can be corrected by bypassing the digraph check on the ts, which is done like so: tat|sulok.


ochakochatblanc:
i like this change - signed nootsance

Cadlaxa:
> i like this change - signed nootsance

😭😭

PR #2151, opened 2026-05-23
Improve the menu name for the default phonemizer

maiko3tattun:
image

When the default phonemizer is selected, it was difficult to tell what this menu represented. This PR explicitly states that “the phonemizer is the default.”


stakira:
Does not feel like an improvement to be honest...

maiko3tattun:
Many people who report that no sound is coming out are actually using the default phonemizer. We tell them to “change the phonemizer.” However, since the track header only displays “DEFAULT,” they don't know where to click.

Alternatively, we can set the default phonemizer for the track to “None” and label it “Select Phonemizer” just like the singer. Do you prefer this approach?

stakira:
Applied and pushed to master in 30eab3ae.

stakira:
(Correction: history was rebased after that comment — the commit is now 468939a6, not the hash previously mentioned.)

PR #2179, opened 2026-06-03
Auto-resize overlapping notes

maiko3tattun:
## Development Background:
- When creating notes, they may overlap and fail to render
- When the Alt key is held down, adjacent notes should be resized; however, if any note is selected, editing of the unselected notes is locked, so they will not be resized

https://github.com/user-attachments/assets/c69d34da-700c-49a2-9eb6-f3c31be63f1a

## Fixes:
- When a note is selected, you can unlock editing for the duration of unselected notes (under the Edit menu in the Piano Roll).
image

- When you add or resize a note, any overlapping notes will be automatically cropped (also in the Edit menu)

https://github.com/user-attachments/assets/6a594501-2e6e-4be0-8ce9-ed1fc91d4f5c

### DIFF
--- a/OpenUtau.Core/Ustx/UNote.cs
+++ b/OpenUtau.Core/Ustx/UNote.cs
@@ -40,9 +40,9 @@ namespace OpenUtau.Core.Ustx {
         [YamlIgnore] public double DurationMs => EndMs - PositionMs;
         [YamlIgnore] public double EndMs { get; set; }
         [YamlIgnore] public bool Selected { get; set; } = false;
-        [YamlIgnore] public UNote Prev { get; set; }
-        [YamlIgnore] public UNote Next { get; set; }
-        [YamlIgnore] public UNote Extends { get; set; }
+        [YamlIgnore] public UNote? Prev { get; set; }
+        [YamlIgnore] public UNote? Next { get; set; }
+        [YamlIgnore] public UNote? Extends { get; set; }
         [YamlIgnore] public int ExtendedDuration { get; set; }
         [YamlIgnore] public int ExtendedEnd => position + ExtendedDuration;
         [YamlIgnore] public int LeftBound => position;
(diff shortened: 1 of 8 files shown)


stakira:
1. I remember we had sth like this when we can only change length from note end. Could you confirm there are not duplicate feature?
2. I think we could change the default behaviour. There seems no reason to keep old behaviour, since you almost certainly will resize the other overlapping note. Is that correct? The benefit is: with less logic branches we can reduce future bugs.

maiko3tattun:
Actually, I thought so too! I didn't understand the purpose of the old specifications, so I included the option just in case, but I also think it's unnecessary.

maiko3tattun:
- Remove the two options and make them always enabled
- For safety, make UNote.Next/Prev/Extend nullable
- Correctly cut overlapping notes even if the note is detached when PointerPressed
- When you revert a note's resize, the neighboring note will also revert
- It works correctly even if the neighboring note isn't snapped to the grid

https://github.com/user-attachments/assets/689c1bfa-2ada-4fcd-8385-86834cbbdf31

maiko3tattun:
As far as I can tell, the only similar feature is batch editing, and this doesn't conflict with drag events.

rokujyushi (changes requested):
キー入力については、引数ではなくメンバー変数を利用して下さい。
[NoteEditState](https://github.com/openutau/OpenUtau/pull/2179/changes#diff-5e873b045729574caad41ef00394752b22d483a12e60f5f69124f6e0a2ce12c7L44)にはキー入力やその他のフラグを受け入れるためのメンバー変数があります。

Regarding key input, please use member variables instead of arguments.
[NoteEditState](https://github.com/openutau/OpenUtau/pull/2179/changes#diff-5e873b045729574caad41ef00394752b22d483a12e60f5f69124f6e0a2ce12c7L44) has member variables for accepting key input and other flags.

stakira (approved):

PR #2182, opened 2026-06-04
Show optional DiffSinger render phrase boundaries in piano roll

KakaruHayate:
## Summary

This PR adds an optional visual overlay for DiffSinger render phrases in the piano roll.

When enabled, OpenUtau draws dashed vertical boundary markers for each DiffSinger render phrase, using the renderer layout bounds so the displayed range matches the actual rendered segment. The overlay is only shown when the active renderer is DiffSinger.

A new DiffSinger preference is added for this behavior:

- `Show render phrase boundaries`
- Default: off

## Testing

- Built successfully with `dotnet build OpenUtau.sln --nologo --verbosity:minimal`
- Launched the OpenUtau desktop app from the debug build


nagotown:
Very nice! Would it be possible to improve the visibility? It's a little difficult to read since the quantize lines are also dashed.

KakaruHayate:
> Very nice! Would it be possible to improve the visibility? It's a little difficult to read since the quantize lines are also dashed.

Thanks for the feedback! I switched the phrase boundary lines to a [4,2,1,2] dot-dash pattern at the full accent color (no longer semi-transparent), so they now read clearly against the [2,4] short-dash quantize grid. Pushed in cdbc5b87.

nagotown:
Another suggestion: I'm not sure how difficult this would be to implement, but it would be easier to immediately recognize if phrases are overlapping if if the overlapping region was made a different color. 

Image below for clarification:

3QmkDoCiRW

KakaruHayate:
> Another suggestion: I'm not sure how difficult this would be to implement, but it would be easier to immediately recognize if phrases are overlapping if if the overlapping region was made a different color.
> 
> Image below for clarification:
> 
> 3QmkDoCiRW

image
okey?

nagotown:
Is there a similar color that's themeable?
I could see this being an issue if someone has a yellow-tinted light theme.

KakaruHayate:
> Is there a similar color that's themeable? I could see this being an issue if someone has a yellow-tinted light theme.

What ideas do you have? You can submit them to my branch.

In fact, I just hope to solve the problem of occasional behavioral errors caused by small gaps in paragraphs when there is no quantification of notes (which often occurs after conversion from other editor projects). As long as there are boundary lines, it's fine. I haven't considered the case of overlap.

yqzhishen (dismissed):

stakira:
What is this? Why the lines cut cross notes? Shouldn't it only happen in empty place?
And, I'd prefer the visual to be much more subtle. E.g., a gradient rectangle or a "]" shape in the waveform area, with the same color as waveform.
And, you don't need a preference. Just leave it on. We don't need a preference for everything.

KakaruHayate:
> What is this? Why the lines cut cross notes? Shouldn't it only happen in empty place? And, I'd prefer the visual to be much more subtle. E.g., a gradient rectangle or a "]" shape in the waveform area, with the same color as waveform. And, you don't need a preference. Just leave it on. We don't need a preference for everything.

Why not? The padding SP before and after the paragraph in the DiffSinger renderer will certainly intrude into the edge notes. This feature is designed to remind users which notes belong to a paragraph.

And really, shouldn't there be no preference? What if someone doesn't like this visual effect? Or is there another way to indicate to users which notes/phonemes are considered as a paragraph in the DiffSinger renderer?

nagotown:
Would it be possible to also add a toggle for render regions in the piano roll's toolbar?

PR #2185, opened 2026-06-06
Log unhandled exceptions

maiko3tattun:
During testing, I confirmed that unexpected exceptions in the main window can be logged. However, it is highly likely that errors occurring before startup is complete will not be captured.


maiko3tattun:
When I looked closely, I saw it was written in other places as well 🫠

PR #2189, opened 2026-06-07
Migrate from Newtonsoft.Json to System.Text.Json

atouu:
Eliminate the need of using two Json parsers and only use System.Text.Json instead.

- Adds `OpenUtau.Core.Json` (similar to `Yaml`) to store default JsonSerializerOptions and wrapper methods for Deserialize and Serialize.


atouu:
May I ask if you can test if Vogen and Voicevox still works without issues? @rokujyushi

rokujyushi:
It seems that Newtonsoft.Json has a mechanism that automatically performs type conversion, but thanks to the strict conversion process of Json.Deserialize<T>, a bug that had remained undetected until now has been discovered.

Click here for the [fix PR](https://github.com/atouu/OpenUtau/pull/5).

rokujyushi (approved):

rokujyushi:
Sorry, I just remembered there was this fix too. My feature addition is probably causing a conflict, so could you please fix it?

PR #2205, opened 2026-06-16
Skip IObservable first init in PreferencesViewModel

atouu:
When opening PreferencesDialog, preferences get resave 43 times (including writing to prefs.json) causing the preferences window to lag for a bit. We can prevent this by adding .Skip(1) to every IObservable.


stakira:
That's a really good finding. The PR is conflict now.

atouu:
conflicts resolved

stakira:
uhh.. conflict again

stakira:
The `.Skip(1)` fix itself is correct and the standard ReactiveUI idiom for this (`WhenAnyValue` replays the current value on subscribe, which is what causes the resave-on-open). Two suggestions to make it more robust against regressing as the file evolves:

1. **Repeating `.Skip(1)` at ~40 call sites is easy to forget for future preferences.** A small helper collapses the duplication and bakes the fix in so it can't be forgotten:
   ```csharp
   void PersistOn<T>(IObservable<T> source, Action<T> assign) {
       source.Skip(1).Subscribe(v => { assign(v); Preferences.Save(); });
   }
   ```
   Then each site is one line, e.g. `PersistOn(this.WhenAnyValue(vm => vm.PreRender), v => Preferences.Default.PreRender = v);`

2. **`ShowOnnxGpu`/`ThemeEditable`/`HighThreads` reinvent a pattern this codebase already has.** Turning them into computed properties with a manual `this.RaisePropertyChanged(nameof(ShowOnnxGpu))` tucked inside the unrelated `OnnxRunner` persistence subscription works, but couples two concerns and is easy to miss if the derivation changes later. `ObservableAsPropertyHelper<T>` + `.ToProperty(...)` (already used in `ExpressionsViewModel`, `MixFxViewModel`, `NotesViewModel`) gives automatic change notification at the property's own declaration site instead:
   ```csharp
   this.WhenAnyValue(vm => vm.OnnxRunner)
       .Select(r => r == "DirectML" || r == "CUDA")
       .ToProperty(this, vm => vm.ShowOnnxGpu, out showOnnxGpu);
   ```

I applied both of these locally against current `master` (which has diverged a bit from this PR's base — extra prefs like `MetronomeVolume`/`Wayland`/`DiffSingerLocalRetaking`/CUDA support) to confirm it builds and the approach holds up: 820928a5. Happy to open it as a PR if useful, or fold the ideas into this one.

atouu:
Another viable solution is do `[Reactive] private int _myVar;` instead of `[Reactive] public partial int MyVar { get; set; }`. And initialize the value in constructor with `_myVar = Preference.Default.MyVar;` which removes the reliance on Skip(1). Only problem here is the code will look ugly.

PR #2217, opened 2026-06-19
Tweaking the curve

rokujyushi:
New curve editing tools have been added to the piano roll.
The following tools have been newly added to the piano roll toolbar:

- Vertical Stretch Tool
- Horizontal Stretch Tool
- Vertical Shift Tool
- Horizontal Shift Tool

Icons and tooltips for these tools have been added, and the settings in PianoRoll.axaml, Strings.axaml, and PianoRollStyles.axaml have been updated accordingly.

Additionally, the behavior of the new tools has been implemented in PianoRoll.axaml.cs, with extensions to the ExpCanvasPointerPressed and ExpCanvasPointerMoved methods.

In CurveViewModel.cs, logic to support the new tools has been added.
The TryGetSelection method has been implemented to retrieve the selection range.

In NoteEditStates.cs, new state classes have been added, and the editing logic (movement and scaling) for each tool has been implemented.

An option to turn off the default snap function for curve selection has been added to Preferences.cs.

Previously, this could be disabled by turning off the Note snap function. However, since some users may not require snapping during selection, a separate setting has been added to allow it to be turned off individually.
The snap function was originally introduced because the selection feature was designed to snap to the grid, matching the specification where curve pasting operations are performed along the grid.


nagotown:
Would it be possible to add a setting to ignore any values that are 0/default when using the shift tools?
I also noticed that selection strictly adheres to the snap division grid, is this intentional?

rokujyushi:
@nagotown 
Regarding the setting to ignore values that are 0 or default when using the shift tools, I'm having a hard time imagining how that would behave.

If you don't want to move the default values, I think it's best to simply not select them.

As for the behavior of the selection, I haven't made any specific changes to it, so it remains exactly as it was when the curve selection feature was first added.

nagotown:
The reason why I wouldn't just select very specific parts is related to the last part of my reply.

It's a bit of a hassle with the grid snapping.
I could just turn off grid snapping, but this means I would have to do it every time I wanted to tweak the pitch.
Drawing pitch doesn't adhere to the grid, so it doesn't make sense to me for it to do that while tweaking the pitch.

By the way, could you check my PR on this branch?
One of the icons is improperly named, causing it to not appear in the expression panel.

rokujyushi:
@nagotown
I have confirmed the grid snapping specifications with the developer, Maiko.
She explained that because the pasting behavior aligns with the grid, the selection process is also designed to snap to it.
Since this behavior is logical, I believe that snapping should be the default setting.
In addition, we will proceed with modifications to allow users to change the default snapping settings.

stakira:
conflict

PR #2218, opened 2026-06-20
[WIP]Add status bar and organize usage tips

maiko3tattun:
## Background of the Development:
- The tooltips for the editing tools contained lengthy and complex text, and their translations might not have been accurate or up-to-date.
- Line breaks in the tooltips were often incorrectly formatted automatically, causing the text to become jumbled.
- There were few ways for users to learn about complex operations involving modifier keys.

## Fixes:
- Internal changes: the tooltip descriptions were broken down into single lines to make them easier to manage.
  image
- Fixed an issue where tooltips were not updating while using Pen Plus Tool

## New Features:
- Added a text area at the bottom to explain operations
  image
- Added tooltips to the Curve Edit tools
  image

## Notes:
- Shortcut keys have been temporarily removed from the tooltips; let's address this after #2096 is merged


stakira:
I don't think we should add the status bar text. The tools have been using tooltip and it should stay that way uniformly. the status bar text could conflict with rendering status, complicating things.

PR #2225, opened 2026-06-22
Chinese VCV phonemizer BY樗儿 support

27704534:
Add Chinese VCV phonemizer support. Produces diphone aliases based on the previous note's tail vowel, following the same pattern as the Japanese VCV phonemizer. Supports hanzi input with automatic romanization via BaseChinesePhonemizer.


lottev1991:
For the record, while the name is non-indicative, the Chinese CVVC phonemizer supports VCV. It's reliant on custom presamp.ini files, meaning it will adapt to the loaded voicebank. This approach is too hardcoded, especially since the Japanese VCV phonemizer is obsolete. I don't think this phonemizer is useful.

27704534:
No，the Chinese CVVC phonemizer didn't supports VCV,You're wrong.
不，中文CVVC不支持VCV
你应该是看到发声了，就觉得工作正常
image
但实际上，音素是错的，全部识别为起始音
image
正常应该是这样

其二，中文VCV录音工程过于庞大，全世界就没有几个中文VCV声库，据我所知，第一个提出的且只有的就是这个形式，就是由“樗儿”提出，所以它的形式十分固定，完全不需要灵活

三，你是对的，世界上公开中文VCV声库寥寥无几，作为这个语素器的开发者，我也仅收集到两个此种声库，以及相关文档，所以，这个语素器确实意义不大。当然因为意义不大，没有多少人使用，我也不那么上心，借鉴了日文VCV语素器。

27704534:
No, the Chinese CVVC phonemizer does not support VCV. You're wrong.
You probably assumed it was working properly just because some sound was produced.

image

In fact, the phonemes are incorrect, as all of them are recognized as initial phonemes.

image

This is what normal results should look like.

Second, Chinese VCV recording projects are extremely large in scale, and there are very few Chinese VCV voicebanks available worldwide. As far as I know, this format was first proposed and remains the only one of its kind, put forward by 樗儿. Its structure is highly fixed, leaving no need for flexible adaptation.

Third, you are right that publicly available Chinese VCV voicebanks are extremely scarce. As the developer of this phonemizer, I have only managed to collect two such voicebanks along with their relevant documentation. Therefore, this phonemizer indeed holds limited practical value. Since it is rarely used due to its low practicality, I did not devote much effort to its development and referenced the Japanese VCV phonemizer in its implementation.

27704534:
[oto模板参考.txt](https://github.com/user-attachments/files/29292997/oto.txt)
[readme.txt](https://github.com/user-attachments/files/29292998/readme.txt)
[Tsuro's_Chinese_7mora_VCV_reclist.txt](https://github.com/user-attachments/files/29293000/Tsuro.s_Chinese_7mora_VCV_reclist.txt)
[每个音与音尾的对应关系.txt](https://github.com/user-attachments/files/29293001/default.txt)
[收录读音.txt](https://github.com/user-attachments/files/29293003/default.txt)
[oto模板.txt](https://github.com/user-attachments/files/29293006/oto.txt)
这里是相关文件

Slidingwall:
It appears that `presamp.ini` fails to load, causing the `prevVowel` logic to break in the ZH CVVC phonemizer.  
Non‑hard‑coded improvements are kinda valuable and can be ported back to the original ZH CVVC phonemizer.

PR #2226, opened 2026-06-22
Improved error message for resampler failures

CdrSonan:
When using the Classic renderer and the resampler encounters an issue, it quietly fails instead of throwing an exception.
An exception is only raised when the wav tool is run in the next step, and it cannot find the resampler output.

This is problematic because the resulting error message contains no information about the actual root cause of the problem, making it effectively useless. This makes spotting oto.ini errors, which are often the cause of resampler failures, as well as debugging the resampler itself more difficult.

This pull request therefore implements a new type of error message specifically for resampler failures:
- During any calls to `ProcessRunner.Run()`, the stdout and stderr streams, as well as the exit code of the subprocess are accumulated into a string variable, which is returned when the process exits or is terminated. This provides a mechanism for tracking the output of the resampler executable. Resamplers might write error-related messages to stdout instead of stderr, which is why stderr is always included. The existing debug flag still controls whether stdout messages are forwarded to the logger.
- After the `ProcessRunner.Run()` call in `DoResamplerReturnsFile()`, it is checked whether an output file has been generated. This seems to be the most reliably way to check for resampler success.
- If the file has not been generated, a new type of exception (`ResamplerFailedException`) is raised. Its error message incorporates the resampler console output captured by `ProcessRunner.Run()`.
- in `RenderEngine.RenderTracks()`, a new else-if is triggered when at least one `ResamplerFailedException` is included in the exception list created by the renderer. In that case, the message of the generated `MessageCustomizableException` is adjusted to include the information that this is a resampler or oto.ini issue, and that generating a singer error report could be useful.

A possible extension of this would be to replace the `MessageCustomizableException` with a new `AggregateResamplerFailedException` class. This could be detected in the `MainWindow`, and would allow displaying a fully custom error popup, including a button to immediately generate a singer error report without navigating through menus. I can implement this as well, so please let me know if you think this is a good idea.


stakira:
Resampler used to pop error dialog. It was considered too annoying and suppressed. Adding it back will be considered too annoying again by people and suppressed again. I'd consider it enough to log the error to logs. You can open the log window to debug.

CdrSonan:
This implementation does not change the number of error dialogues the user sees.

It replaces the aggregate exception error dialogue containing the wavtool exception(s) with a new, different aggregate exception dialogue that's hopefully more useful.

You can see this when you look at the diff in RenderEngine.cs. The new case for resampler exceptions mirrors the default case these errors would previously fall under. Only the error message is changed.

PR #2227, opened 2026-06-24
Fix Piano Roll Window Mode Not Reset When Switching from Detach to Hide

27704534:
When the user switches the Piano Roll window mode from Detach to Hide, the DetachPianoRoll preference is not reset to false. This causes the Piano Roll to remain in detached mode when it is shown again, instead of embedding back into the main window as expected.

### DIFF
--- a/OpenUtau/Controls/PianoRoll.axaml
+++ b/OpenUtau/Controls/PianoRoll.axaml
@@ -359,7 +359,7 @@
                     <CheckBox IsChecked="{Binding PianoRollDetached}" />
                   </MenuItem.Icon>
                 </MenuItem>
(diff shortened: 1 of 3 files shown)


atouu:
Hiding it is the expected behavior, it works the same as closing the Piano Roll Window. I think it's better to just set the MenuItem to IsVisible=false when detached, now that I'm looking at it.

27704534:
我理解错了hide的意思
这个hide原来是关闭钢琴卷帘的意思，我以为是隐藏独立的窗口，改为内嵌，还在纳闷怎么这个功能怎么这么怪
既然这样，那确实够繁琐的，分离模式下不需要它，确实需要隐藏。内嵌模式下它竟然藏在三级菜单，一点也不便利

atouu:
Yeah, unfortunately I can't figure out where to put the hide button to make it easier to access.

27704534:
I will do it for you😁

PR #2228, opened 2026-06-24
Nightly Builds

atouu:
## Why?
Because only having beta and stable makes it harder for us contributors to find bugs and fix them. Having nightly builds allow more people to test features as soon as possible and also find bugs that we can fix right away

## How it works
1. Nightly checks for code changes from these directories every 01:00 UTC:
    - `OpenUtau`
    - `OpenUtau.Core`
    - `OpenUtau.Plugin.Builtin`
    - `runtimes`
2. Then triggers the build workflow and release the artifacts in `nightly` release tag (this release will be overwritten every build).
> Installer builds are disabled in nightly to avoid conflicts with stable and beta releases

## Creating the `nightly` tag.

To avoid picking up nightly as latest, we can create an empty nightly tag with specified date so release can use it instead of the current date. This tag needs to be created before running the nightly workflow.

1. Create an orphan branch
```sh
git switch --orphan empty
git clean -fdx
```
2. Create a README.md notice (recommended since Github does not allow removing the auto-generated `.tar.gz` and `.zip` source code)
```
# OpenUtau Nightly

No source code here, please checkout the `master` branch
```
3. Create and push the tag
```bash
git add README.md
git commit -m "nightly"
GIT_COMMITTER_DATE="2024-08-01T12:00:00" git tag -a nightly -m nightly
git push origin nightly
```
4. The `empty` branch can be safely remove after push.

The release will be displayed before `0.1.539 Beta`.

## Preview

Screenshot_2026-06-25_01-02-00

## Other Changes
- Use Informational Version to display full version string.


ochakochatblanc:
i ackree:3

27704534:
Good，Very useful

ain-soph:
Update on this? It's quite useful.

PR #2230, opened 2026-06-25
Piano Roll UI Improvements: Add toolbar close button, double-click to close, optimize detach mode UXll

27704534:
image
image
image
image

Adds a close button to the top-right corner of the piano roll toolbar, along with a few quality-of-life improvements for detach mode.

New Features

Toolbar Close Button
Added an X button at the top-right corner of the piano roll — click to hide the roll. The button position stays fixed and doesn't shift when the parameter panel toggles, which feels natural to Windows users. It automatically hides in detach mode to avoid duplication with the window's native close button.

Double-Click to Close
Double-clicking the empty area of the toolbar also closes the piano roll. Threshold is set to 300ms to prevent accidental triggers.

Detach Mode Menu Polish
The "View → Piano Roll → Hide" menu item is now automatically hidden in detach mode, since the window already has a system close button making it redundant.
Implementation Notes

Uses a "floating button + placeholder" layout approach: the close button stays fixed at the top-right corner, and a placeholder element shifts the parameter panel button left by 20px when the panel is collapsed to prevent overlap. The shift is barely noticeable in normal use, but the fixed close button position is a clear quality-of-life improvement.

The close button reuses the existing toolbar ToggleButton style. IsChecked is manually reset after click to avoid leftover selected background on the next open — the user never sees this since the roll hides immediately.

The placeholder uses nested double Border for multi-condition visibility (detach mode + panel state) without adding new ViewModel properties, keeping changes minimal.

Files Changed

OpenUtau/Controls/PianoRoll.axaml 
OpenUtau/ViewModels/PianoRollViewModel.cs 
OpenUtau/Controls/PianoRoll.axaml.cs 

Compatibility

All changes are fully backward compatible and do not affect core functionality like rendering or editing.


27704534:
Because the China's bad internet,And my Cloudflare WARP didn't work well.So my Git and Nuget didn't work well.I have to use the strange way to edit it.

atouu:
Hello, I opened a PR in https://github.com/27704534/OpenUtau/pull/1 that improves this:

> Changes:
> - Use standard button instead of toggle for hide
> - Use DoubleTapped instead
> - Restructure menus and strings
> - Merged [detach-pianoroll-prefs-toggle](https://github.com/atouu/OpenUtau/tree/detach-pianoroll-prefs-toggle) to here.

|  |  |
| --- | --- |



Also, I recommend not to over comment if the code is already clear.

atouu (approved):

PR #2231, opened 2026-06-25
support-svp from PR#1525

27704534:
From:https://github.com/openutau/OpenUtau/pull/1525
Thank Cadlaxa very much.


27704534:
Fixes multiple issues when importing SVP (Synthesizer V project) files, improving import compatibility and completeness.

Fixed Issues
Null Reference Crash
Missing AfterLoad() call after SVP import leaves Part objects semi-initialized, triggering NullReferenceException when opening the project.

Fix: Add project.AfterLoad() before ValidateFull() at the end of ConvertToUstx in both SVP and SVP2.
Loose Notes Loss
"Loose notes" written directly on the piano roll (without manual grouping) are stored in mainGroup. The original code only processed groups with SVPMRef references, causing all loose notes to be lost.

Fix: Detect if mainGroup contains notes, assign a uuid and add to libraryGroups, create a default reference and insert it at the head of the allRefs list to reuse existing group processing logic.
Missing groups Field in SV1 Format
SVP (SV1 format) only processed the library field, not the groups field, causing groups in some projects to fail loading.

Fix: Add iteration handling for svpProject.groups in the libraryGroups initialization section.
SV1 Timeline Boundary Checks
timeSignatures and tempos in SVP (SV1) lacked boundary checks, potentially causing negative indices.

Fix: Add Math.Max(0, ...) boundary checks, aligned with SVP2 format.
Relaxed mainGroup Fallback Logic
The original code only used mainGroup as fallback when groupID matched or was empty, which was too strict and caused some projects to fail finding the corresponding group.

Fix: Use mainGroup as fallback whenever libraryGroups doesn't find the group and mainGroup is not null.
Track Volume & Pan Conversion
The original code didn't convert track volume, pan, mute, or solo parameters at all, leaving all tracks at default settings after import.

Fix:
Add SVPMixer class to parse the mixer field in SVP files
Volume: Directly assign dB value, clamp above +12 to +12
Pan: Convert from -1~1 range to -100~100
Mute: Set track.Mute
Solo: Set track.Solo
Mute Property Name Correction
The original code incorrectly used Muted (runtime state, YamlIgnore). The correct persisted property is Mute.

Fix: Use track.Mute instead of track.Muted.
Files Changed
OpenUtau.Core/Format/SVP.cs
Compatibility
All changes are import-only fixes and do not affect existing project editing or rendering. Fully backward compatible.

I use AI to translate

nagotown:
@27704534 Why not PR to the original PR's branch?

27704534:
Sorry,The current version has way too many issues.I'll submit it to the repo after I finish polishing it.I messed up

PR #2238, opened 2026-06-27
Add More Track Effects + Fixes

Cadlaxa:
- Adds De-esser, De-Thumper and Saturation
- Fix sliders to reflect the actual applied effect when reopening the fx window
- Add right click to reset the sliders

### DIFF
--- a/OpenUtau.Core/SignalChain/Effects/DeEsser.cs
+++ b/OpenUtau.Core/SignalChain/Effects/DeEsser.cs
@@ -1,0 +1,75 @@
+﻿using System;
+
+namespace OpenUtau.Core.SignalChain.Effects {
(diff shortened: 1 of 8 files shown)


Cadlaxa:
image

KakaruHayate:
I think this is inappropriate. These effects are often applied locally. And I can't think of why we need to deal with these issues in a singing voice synthesis engine. I added the effector before to benchmark against the current mainstream singing voice synthesis software during audition (they have updated similar functions), but are more functions crossing the line? If this continues, it will become increasingly bloated

Cadlaxa:
Yeah you're right

nagotown:
would you mind making a new pr for the improvements other than the new effects?

PR #2241, opened 2026-06-30
Fix pitch line tool icon used in expression window

nagotown:
Not currently visible, but is required to display properly for #2217.
Changes `ListBoxItem.pitchLineTool` to `ListBoxItem.drawLinePitchTool`.
For whatever reason, that PR does not change this when merging despite the line being clearly different compared to main.
[master](https://github.com/rokujyushi/OpenUtau/blob/7224f4cc2754f1752c33e7a27cd9ec6e11d110f3/OpenUtau/Styles/PianoRollStyles.axaml#L152) / [#2217](https://github.com/openutau/OpenUtau/blob/master/OpenUtau/Styles/PianoRollStyles.axaml#L137)


nagotown:
Got confused here. The master branch never uses `drawLinePitchTool`, which means it was only renamed in the PR.
Still weird that it doesn't register in the diff though.

PR #2259, opened 2026-07-11
Fix Phonemizer ALT handling

maiko3tattun:
Fix #2236's bug


BagelHero:
As mentioned in the discord: 
Requires reversion of #2284 to merge this instead. This is a more fundamental solution that solves the bug more broadly instead of on a per-phonemizer basis

**This needs to be merged before other phonemizer-related PRs.** It's causing downstream issues, I believe, so this is high priority.

maiko3tattun:
I've included the revert from #2284 in this PR. You can just merge it now.

PR #2265, opened 2026-07-15
Fix UstFlagParser bug

maiko3tattun:
Fixed a bug where the “/” flag for resampler switching was not parsed


maiko3tattun:
The test code assumed that `Parse()` would not return the slash flag. Now that `Parse()` returns the slash flag as well, users are notified that it was not imported from ust file.

PR #2183, opened 2026-06-05
Support DiffSinger pitch local retaking

KakaruHayate:

PR #2287, opened 2026-08-04
[DiffSinger/BUGFIX]Implement DiffSinger variance retake with hard compose

KakaruHayate:
## Merge order and dependencies

1. Merge or rebase #2183 first. It provides the DiffSinger pitch retake frame-mask and non-contiguous retake write-back foundation.
2. Merge this variance retake change next. It replaces the variance crossfade and cumulative baseline behavior introduced by #2176 with tensor-cache-based partial retake and hard composition.
3. Rebase or merge #2175 after this change so the real-curve scheduler consumes the final hard-composed variance result.
4. Process #2174 last. Its render-priority changes are ordering-only and should be reconciled with #2175 changes in `RenderEngine.cs` and `Notifications.cs`.

This PR is intentionally based on current `master`, where #2176 is already merged. It does not include the code from #2183, #2175, or #2174. The order above describes the intended final integration order; if this PR is reviewed before #2183 is merged, it can remain independent on `master` and be rebased when the stack is integrated.

## Retake scope requirements

Partial variance retake is allowed only when the tensor-cache scope remains compatible:

- Pitch and frame-level speaker embedding changes are retake-eligible.
- Linguistic encoder inputs or outputs, phoneme durations, word durations/divisions, phrase frame layout, model identity, variance channel layout, and diffusion steps/speedup remain part of the scope key. A change to any of them uses a different scope and performs full generation.
- Tensor cache must be enabled and a compatible previous result must exist.
- The first version does not expand the frame mask.
- No crossfade is used.
- Model output outside the mask is not trusted. Hard composition restores every unmasked frame from the previous cached result.

## What changed

- Split complete variance cache lookup from the partial-retake scope.
- Reuse cached variance curves as predictor inputs during partial retake.
- Build a frame mask from pitch and speaker-embedding changes, then expand it to the model's `[1, T, V]` retake input.
- Hard-compose predicted frames with the previous result after inference.
- Save the final composed result to the complete tensor cache so undo/redo can reuse exact historical states.
- Remove the fixed 50 ms crossfade, weighted blending, and cumulative merged baseline.
- Add focused tests for frame masks, frame-grouped speaker embeddings, channel expansion, hard composition, leakage prevention, and incompatible metadata.

## Validation

- `dotnet restore OpenUtau.Test/OpenUtau.Test.csproj`
- `dotnet test OpenUtau.Test/OpenUtau.Test.csproj --no-restore --filter "FullyQualifiedName~DiffSingerVariancePatchTest"`


stakira:
This PR flips line endings.

PR #2363, opened 2026-09-04
Fix AutoConvel Track.part bug

Cadlaxa:
Fix autoConvel desync when moving parts by accounting for the part's position offset in note lookups and BPM sampling.


ochakochatblanc (approved):
LGTM

OpenUtau.Plugin.Builtin/EnglishVCCVPhonemizer.cs
```diff
@@ -70,6 +70,7 @@ public EnglishVCCVPhonemizer() {
                 {"ir","Er"},
                 {"ir-","Er-"},
                 {"6 r","or-"},
+                {"6r-","0r-"},
```
AnAndroNerd, line 73: should technically be [or-] but using replacements in the YAML is better

AnAndroNerd (approved):

PR #2370, opened 2026-09-06
[Beta Bug] Fix EN-XSAMPA Legacy ValidateAlias phonemes

Cadlaxa:
this fixes the old legacy fallback lotteV implemented before


nagotown (approved):
works on my end 👍

stakira (approved):

PR #2376, opened 2026-09-06
Wire the DAW integration API into the UI

KakaruHayate:
## Summary
Follow-up to #2369: exposes the DAW integration core in the UI so the feature is actually usable.

- Adds a **Tools > DAW Integration** menu entry that opens a connection dialog.
- The dialog lists discovered plugin servers (from the discovery directory), connects/disconnects them, and reflects `DawManager` state (connecting / connected count / automatic reconnecting / connection lost).
- Strings: 25 `dawintegration.*` keys + `menu.tools.dawintegration`; also restores `button.close` which had been removed.
- View models use `ReactiveUI.SourceGenerators` (`[Reactive] public partial` properties) per the current baseline style.
- Connect/Disconnect gating is derived from manager connection state: Connect only while the selected server is disconnected, Disconnect enabled while any manager connection is live (including reconnecting), and teardown targets the selected port only when it actually has a connection.

## Notes
- No changes to `OpenUtau.Core`; UI-layer only.
- The same change was reviewed and closed out on my fork (KakaruHayate/OpenUtau#16, CodeRabbit review addressed in a9878199); this branch is a cherry-pick of those two commits onto current master.

## Verification
- `dotnet build OpenUtau/OpenUtau.csproj` - 0 errors.
- `dotnet test --filter FullyQualifiedName~DawIntegration` - 83 passed, 1 skipped.

### DIFF
--- a/OpenUtau/Views/DawIntegrationDialog.axaml
+++ b/OpenUtau/Views/DawIntegrationDialog.axaml
@@ -1,0 +1,44 @@
+<Window xmlns="https://github.com/avaloniaui"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        mc:Ignorable="d" Width="680" Height="400" MinWidth="520" MinHeight="320"
+        x:Class="OpenUtau.App.Views.DawIntegrationDialog"
+        Icon="/Assets/open-utau.ico"
+        Title="{DynamicResource dawintegration.caption}"
+        WindowStartupLocation="CenterOwner">
+  <Grid Margin="10" RowDefinitions="Auto,1*,Auto,Auto">
+    <TextBlock Grid.Row="0" Text="{DynamicResource dawintegration.hint}"
+               TextWrapping="Wrap" Margin="0,0,0,8"/>
+
+    <DataGrid Grid.Row="1" ItemsSource="{Binding Servers}" SelectedItem="{Binding SelectedServer}"
+              AutoGenerateColumns="False" CanUserSortColumns="False" IsReadOnly="True"
+              BorderThickness="1" BorderBrush="{DynamicResource SystemControlHighlightBaseMediumHighBrush}">
+      <DataGrid.Columns>
+        <DataGridTextColumn Header="{DynamicResource dawintegration.col.name}"
+                            Binding="{Binding Name}" Width="*"/>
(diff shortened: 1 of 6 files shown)


OpenUtau/Views/DawIntegrationDialog.axaml
```diff
@@ -0,0 +1,44 @@
...
+
+    <StackPanel Grid.Row="3" Orientation="Horizontal" Spacing="10" Margin="0,10,0,0"
+                HorizontalAlignment="Right">
+      <Button Content="{DynamicResource dawintegration.refresh}" Click="OnRefresh"
+              IsEnabled="{Binding !IsBusy}"/>
+      <Button Content="{DynamicResource dawintegration.connect}" Click="OnConnect"
+              IsEnabled="{Binding ConnectEnabled}"/>
+      <Button Content="{DynamicResource dawintegration.disconnect}" Click="OnDisconnect"
+              IsEnabled="{Binding DisconnectEnabled}"/>
+      <Button Content="{DynamicResource button.close}" Click="OnClose" Width="90"/>
```
rokujyushi: https://github.com/openutau/OpenUtau/pull/2376/changes#diff-1ddf8eb586ec74fc554aabbb9aa2e4ee4570bfb845dd0b2c0b5bb1d852306f76R7

If the goal is simply to close the screen and maintain consistency in behavior and appearance with other windows, having only the window's close button (the X button) would be sufficient.

rokujyushi: If there is a specific reason for implementing an individual close button, I think it would also be helpful for designing future windows.

OpenUtau/Views/DawIntegrationDialog.axaml
```diff
@@ -0,0 +1,44 @@
...
+    <DataGrid Grid.Row="1" ItemsSource="{Binding Servers}" SelectedItem="{Binding SelectedServer}"
+              AutoGenerateColumns="False" CanUserSortColumns="False" IsReadOnly="True"
+              BorderThickness="1" BorderBrush="{DynamicResource SystemControlHighlightBaseMediumHighBrush}">
+      <DataGrid.Columns>
+        <DataGridTextColumn Header="{DynamicResource dawintegration.col.name}"
+                            Binding="{Binding Name}" Width="*"/>
+        <DataGridTextColumn Header="{DynamicResource dawintegration.col.port}"
+                            Binding="{Binding Port}" Width="70"/>
+        <DataGridTextColumn Header="{DynamicResource dawintegration.col.apiversion}"
+                            Binding="{Binding ApiVersion}" Width="60"/>
```
rokujyushi: Image. 
Depending on the font used, the text may become truncated. Therefore, we would appreciate it if you could adjust the width or consider making the width adjustable.

PR #2400, opened 2026-09-12
Pre-Roll and Auto Move Playhead

maiko3tattun:
## New Features:
- Auto Move Playhead: When notes are selected, the playhead moves to the position of the first note in the selection. This feature is disabled by default.
  image   
- Pre-roll: Playback starts a specified number of measures before the playhead’s current position. This feature is disabled by default; you can change the number of measures by right-clicking the button in the MainWindow.
  image   
  image   
- Both of these are common features in DAWs.

## Internal Changes:
- The code that moves the playhead when playback stops has been moved from the ViewModel to the Model.
- A generic ComboBoxDialog has been added.


stakira:
We can select playback range now. I don't see why we need to determine playback position from selected notes.

maiko3tattun:
The reason I want to move the playhead is simply because moving it with a click is cumbersome. While editing, you end up moving the mouse cursor far away from the notes repeatedly, and over time, this adds up to a significant loss of time. Also, everyone has probably experienced accidentally creating a note while trying to click on the timeline.
This issue isn’t resolved by range playback. In a workflow where you start playback from the beginning, edit the parts you’re interested in, and then resume playback, this is the optimal solution.

PR #2407, opened 2026-09-13
Fix ReadDictionaryAndInit, Parent Alt, Voice color, and ToneShift

Cadlaxa:
Switching singers caused a race condition where the phonemizer tried to look up notes before the voicebank and its dictionary finished reloading. Because the singer was not ready, valid phonemes were errored out


stakira:
Broad silent try catch hides real problem instead of fixing them. It also hides problem in the future, making debugging difficult.

stakira:
> Broad silent try catch hides real problem instead of fixing them. It also hides problem in the future, making debugging difficult.

This is about changes in OpenUtau.Core/Api/Phonemizer.cs.

Issue #2417, opened 2026-09-16
Multi-selection of notes with ctrl+lmb not working properly


nagotown:
### Acknowledgement

- [x] I have read [Getting-Started](https://github.com/stakira/OpenUtau/wiki/Getting-Started) and [FAQ](https://github.com/stakira/OpenUtau/wiki/FAQ)
- [x] If dealing with version 0.1.549 or higher, you have installed the .NET 8 runtime

### 🐛 Describe the bug

Multi-selection of notes is broken right now. 
Not sure how many versions or commits this has lasted for, but it's happening when built on the latest commit and as been for several of them.

When attempting to select multiple notes via ctrl + click, the previously selected note will be unselected.

<img width="1495" height="1005" alt="Image" src="https://github.com/user-attachments/assets/9128f901-4cb8-49f0-a7bf-1431ba366245" /><br>

There is an exception where if the second selected note is selected via dragging, the rest can be selected with a click.

<img width="2242" height="1135" alt="Image" src="https://github.com/user-attachments/assets/60080023-0c18-43fc-8be1-cfed0c88486a" /><br>

It works fine with parts on the main window/track view, maybe because clicking normally with ctrl held triggers a hold state even if not actually holding (visual indication).

<img width="1167" height="702" alt="Image" src="https://github.com/user-attachments/assets/019383e7-c891-46f6-be9b-f11496a2346b" />

### Explains how to reproduce the bug

1. Hold ctrl + click multiple notes
2. Previous note will be unselected

### OS & Version

Windows 10 / OpenUtau built on commit 81637a33ceeaedd8c407a268ad048e182592508e

### Logs

[log20260916.txt](https://github.com/user-attachments/files/32278388/log20260916.txt)

<details>
<summary>Click to open log</summary>

```
2026-09-16 01:21:21.931 -07:00 [INF] Logging initialized.
2026-09-16 01:21:21.962 -07:00 [INF] Microsoft Windows NT 10.0.19045.0
2026-09-16 01:21:21.963 -07:00 [INF] Microsoft Windows 10.0.19045 X64 X64
2026-09-16 01:21:21.963 -07:00 [INF] OpenUtau v0.0.0.0 win-x64
2026-09-16 01:21:21.963 -07:00 [INF] Data path = E:\Github\OUtesting\OpenUtau\OpenUtau\bin\Debug\ou_newmain_multiselerr
[...]

PR #2419, opened 2026-09-16
Fix Ctrl+LMB Note Selection & Only Show LyricBox if KeyModifiers is None

atouu:
https://github.com/user-attachments/assets/aa8037f6-1c8a-48a5-9a34-274fd951a3d2


atouu:
Fixes #2417

PR #2420, opened 2026-09-17
Fix Testing in the Build.yml

AnAndroNerd:


rokujyushi (approved):
We have verified that the command can be executed in the local environment and have checked the validity of the command.
Since build support may not be fully migrated in IDE environments other than Visual Studio, we have limited this to a provisional workaround for the project execution command, rather than executing tests.

PR #2428, opened 2026-09-21
[SBP] Fix Case-Sensitivity Issue

Cadlaxa:
- Uses StringComparer.Ordinal for everything, from symbols, replacements, fallbacks, ValidateAlias etc
- Implement GetHardcodedFallbacks and ValidateAlias YAML prioritization (phoneme supression from builtin hardcoded fallbacks from child phonemizers)

### DIFF
--- a/OpenUtau.Core/Api/Phonemizer.cs
+++ b/OpenUtau.Core/Api/Phonemizer.cs
@@ -276,8 +276,12 @@ namespace OpenUtau.Api {
 
         public int GetParentToneShift() {
             if (project != null && track != null) {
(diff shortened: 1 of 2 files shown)


OpenUtau.Core/Api/Phonemizer.cs
stakira, line 1: Do you need to change this file?

Cadlaxa: Oh, I didn't touched the phonemizer.cs, Ig it got carried before using this same branch, the fixes are only within the SyllableBasedPhonemizer api

